PR #8, opened 2022-08-30
Chore: Add DefaultView docs, update examples with small explanation

cizl:
Default view docs are missing fixed vs free positioning explanation.


docs/view-default.md
```diff
@@ -0,0 +1,169 @@
...
+All you have to do is return a `IPosition` that requires 2 basic properties: `x` and `y`. (`{ x: node.data.posX, y: node.data.posY }`)
+
+### Property `render`
+
+Optional property `render` has several rendering options that you can tweak. Read more about them
+on [Styling guide](./styles.md).
+
+### Property `simulation`
+
+Fine-grained D3 simulation engine settings. This may be condensed into fewer, more abstract settings in the future.
```
tonilastre, line 115: It would be great to, at least, list the properties that can be updated within `simulation` and a link to D3 documentation to read more about it. The user, if not using Typescript, needs to get into the code to figure out all the properties that can be set up in `simulation` settings.

tonilastre (changes requested):

docs/view-default.md
cizl: I thought about this, but we were talking about maybe abstracting this stuff away into a single property, so it didn't make sense for me to do work that will be deleted.

tonilastre (approved):

PR #11, opened 2022-09-09
Chore: Add pre commit hook

tonilastre:


package.json
```diff
@@ -76,6 +77,7 @@
     "eslint-plugin-jest": "26.2.2",
     "eslint-plugin-prettier": "4.0.0",
     "http-server": "^14.1.1",
+    "husky": "^8.0.1",
```
cizl, line 80: Is it okay to use carets here? (`^`)

tonilastre: I think `devDependencies` is ok because it is a local development mode. It is not included in the release build like `dependencies` is. What do you think?

tonilastre: I will merge this, but let's sync regarding `devDependencies` and if we want to have fixed versions there.

PR #13, opened 2022-09-12
Chore: Fix docs and general orb API issues

tonilastre:
It includes the following changes:

* Examples are more informative: better file naming, expanded description
* README and docs examples are all Typescript first, e.g. `new Orb<MyNode, MyEdge>(container)`
* Removed `get` and `set` from `INode` and `IEdge` interfaces
* Replaced `node.properties` to `node.style` and `edge.properties` to `edge.style`
* Fixed `OrbEventType.NODE_HOVER` to emit only once on node hover event
* Added `DefaultView` and `MapView` settings interface and default settings configuration in the docs


cizl (approved):
Are you using some kind of formatter for markdown? I'm noticing it's cutting off lines at certain column sizes.

We should align our editors here since the formatted will always modify some lines when saving a file. Not an issue at all, but it would be easier to filter out noise when reviewing PRs.

tonilastre:
> Are you using some kind of formatter for markdown? I'm noticing it's cutting off lines at certain column sizes.
> 
> We should align our editors here since the formatted will always modify some lines when saving a file. Not an issue at all, but it would be easier to filter out noise when reviewing PRs.

Hahah I wish I did. It is just OCD :D But I am sure there is a formatter for it, maybe even eslint can do it in regards to the max line length.

Issue #16, opened 2022-09-16
Fix the Typescript usage of the view settings


tonilastre:
Issue: Typescript interface is not correctly applied to view (`DefaultView`, `MapView`) settings

Details can be read on Stackoverflow: [How to setup typescript generics in class constructors and functions](https://stackoverflow.com/questions/73429628/how-to-setup-typescript-generics-in-class-constructors-and-functions)

The problem is that main `orb` object handles different views through `orb.setView` where each view can have different settings so `orb.view.setSettings` is not able to apply correct Typescript interface (at least with the current latest TS version).

The solution could be to create a new instance on `view` changes, e.g.

```typescript
const orb = new Orb({ generalSettings });

const view = orb.createView((context) => new DefaultView(container1, { viewSettings }));

// Use view from now on
view.render(() => {
  view.recenter();
});

const newView = orb.createView((context) => new MapView(container2, { viewSettings }));
```

Issue #23, opened 2022-10-05
Fix collapsing graph container height


cizl:
If the provided graph container doesn't have a defined height or its calculated height equals to 0, the graph isn't visible and it's hard to debug.

tonilastre:
Fixed in release `v0.1.2`.

Issue #22, opened 2022-10-05
Stop removing child elements from graph container.


cizl:
Orb seems to remove all content from the provided graph container. It would make more sense to have it append the canvas to the container instead of replacing its content.

tonilastre:
Fixed in release `v0.1.2`.

PR #24, opened 2022-10-05
Chore: Set dimensions on parent element, Preserve graph container contents

cizl:
* Set dimensions on parent element when undefined (https://github.com/memgraph/orb/issues/23)
* Preserve graph container children (https://github.com/memgraph/orb/issues/22)

### DIFF
--- a/src/utils/html.utils.ts
+++ b/src/utils/html.utils.ts
@@ -1,0 +1,36 @@
+export const setupContainer = (container: HTMLElement) => {
+  container.style.position = 'relative';
+  const style = getComputedStyle(container);
+  if (!style.display) {
+    container.style.display = 'block';
+    console.warn("[Orb] Graph container doesn't have defined 'display' property. Setting 'display' to 'block'...");
+  }
+  if (!style.width || style.width === '0px') {
+    container.style.width = '100%';
+    if (getComputedStyle(container).width === '0px') {
+      container.style.width = '400px';
+      console.warn(
+        "[Orb] The graph container element and its parent don't have defined width properties.",
+        'If you are using percentage values,',
+        'please make sure that the parent element of the graph container has a defined position and width.',
+        "Setting the width of the graph container to an arbirtrary value of '400px'...",
+      );
+    } else {
+      console.warn("[Orb] The graph container element doesn't have defined width. Setting width to 100%...");
+    }
+  }
+  if (!style.height || style.height === '0px') {
+    container.style.height = '100%';
+    if (getComputedStyle(container).height === '0px') {
+      container.style.height = '400px';
+      console.warn(
+        "[Orb] The graph container element and its parent don't have defined height properties.",
+        'If you are using percentage values,',
+        'please make sure that the parent element of the graph container has a defined position and height.',
+        "Setting the height of the graph container to an arbitrary value of '400px'...",
+      );
+    } else {
+      console.warn("[Orb] Graph container doesn't have defined height. Setting height to 100%...");
+    }
+  }
+};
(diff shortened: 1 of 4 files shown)


src/utils/html.utils.ts
```diff
@@ -0,0 +1,36 @@
+export const setupContainer = (container: HTMLElement) => {
+  container.style.position = 'relative';
+  const style = getComputedStyle(container);
+  if (!style.display) {
+    container.style.display = 'block';
+    console.warn("[Orb] Graph container doesn't have defined 'display' property. Setting 'display' to 'block'...");
+  }
+  if (!style.width || style.width === '0px') {
```
tonilastre, line 8: Is this covering cases like `'0rem'`, `'0'`, `'0 px'` if those are possible values for `.width` and `.height` (down below)?

src/views/default-view.ts
```diff
@@ -389,6 +388,17 @@ export class DefaultView<N extends INodeBase, E extends IEdgeBase> implements IO
     }
   };
 
+  private _initCanvas(): HTMLCanvasElement {
+    const canvas = document.createElement('canvas');
+    canvas.id = 'orbCanvas';
```
tonilastre, line 393: What will happen if there is a `HTMLElement` with the same `id` in the application using Orb? Do we want to introduce `uuid` for these kind of ids?

cizl: Good comment, I was just wondering that. I'm assigning `uuid`s in `ngx-orb`, but we may want to have those in here as well. But I'm not sure wether this is the responsibility of the user or the library. It also makes sense to me that the user takes care of his own container and orb instances.

This is effectively a leftover so I'll delete it. I used this to get the canvas element but then i realized that I already have a reference to the element (`this._canvas`)

src/utils/html.utils.ts
cizl: Good point. In firefox the `getComputedSize` returns a `0px` specifically. I can check if it `startsWith('0')` or something like that.

Another point is, what if the user programmatically is switching sizes, then we override that with this condition.

tonilastre: Yep, that should do the trick. I just tried this:

```html
<div id="testing-height" style="height: 010px"></div>
```

```js
document.getElementById("testing-height").style.height // outputs '10px'
```

All in all, `010px` is a valid style, but luckily browser will take out the leading zeros.

Regarding user switching sizes, could we maybe then add settings param for it? Which is enabled by default. What do you think?

cizl: @tonilastre That definitely makes sense! Justin Case - I added a regex to capture all "collapsed" style dimensions.

I've been meaning to talk to you about our settings. I'm not very happy with the current implementation. It's tricky to consume settings from a DX perspective. Plus, now they have duplicated fields. We could add a base interface for ViewSettings in the future.

tonilastre: I agree with you regarding settings. I think we should do that along with the following fix: https://github.com/memgraph/orb/issues/16. With that in mind, we could create a baseline interface as well as `AbstractView`.

src/utils/html.utils.ts
```diff
@@ -0,0 +1,50 @@
...
+        'please make sure that the parent element of the graph container has a defined position and height.',
+        "Setting the height of the graph container to an arbitrary value of '400px'...",
+      );
+    } else {
+      console.warn("[Orb] Graph container doesn't have defined height. Setting height to 100%...");
+    }
+  }
+};
+
+export const collapsedDimensionRegex = /^\s*0+\s*(?:px|rem|em|vh|vw)?\s*$/;
```
tonilastre, line 42: If `px`, `rem`, `em` can be the upper case then you need to add an ignore-case flag at the end of the regex, `/..../i;`

tonilastre (approved):

PR #34, opened 2022-11-07
New: Change the API to handle OrbView and OrbMapView

tonilastre:
The following PR will include API changes to the views such as:

* There are separate classes for orb default view (`OrbView`) and orb map view (`OrbMapView`)
* It fixes the Typescript issue with settings: #16 
* Orb view (`OrbView` or `OrbMapView`) is the main point where it receives a `container` and optional `settings`, `data`, and `events` are part of the orb view
* When changing the views on the same container, `view.destroy()` should be called on the previous view to remove event listeners and content of the container
* Hover and click strategy changed into view parameters (`strategy.isDefaultSelectEnabled`, `strategy.isDefaultHoverEnabled`) with an example in `docs/...md` on how to overwrite with custom select/hover strategy
* Fixed the issue when the strategy was missing, click/hover were not working


cizl (approved):
The API even looks cleaner now, and I like the new strategy handling.

tonilastre:
> The API even looks cleaner now, and I like the new strategy handling.
> 
> It seems to me that are starting to have a couple of reused pieces of markdown in our docs (the strategy section for example). This might get difficult to maintain down the line. Just thinking out loud: maybe we could introduce pandoc or some other preprocessor to build our final docs. (for example reference and reuse the strategy section in multiple files)

You are 100% right. We can touch base on the docs after you are done with the simulator part. Skip the docs changes for it so we can figure out how to handle it. How does that sound to you?

Issue #28, opened 2022-10-11
Fix: Using Orb with Webpack4


KenCarroll:
Exploring how to use Orb in a Vue/Nuxtjs project.  Get the following error:

ERROR in ./node_modules/@memgraph/orb/dist/simulator/types/web-worker-simulator/simulator.js 7:34
Module parse failed: Unexpected token (7:34)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|         this.worker = new Worker(new URL(
|         /* webpackChunkName: 'process.worker' */
>         './process.worker', import.meta.url));
|         this.worker.onmessage = ({ data }) => {
|             var _a, _b, _c, _d, _e, _f;

Researching the problem found this:

- https://substrate.stackexchange.com/questions/855/best-way-to-configure-webpack4-and-polkadot-dependencies-in-a-webworker

which led me to these:
- https://github.com/webpack/webpack/issues/6719
- https://www.npmjs.com/package/@open-wc/webpack-import-meta-loader
- npm i @open-wc/webpack-import-meta-loader

Adding loaders into nuxt.config.js https://nuxtjs.org/docs/features/configuration/#extend-webpack-config
  build: {
    extend(config, { isDev, isClient }) {
      // ..
      config.module.rules.push({
        test: /\.js$/,
        loader: require.resolve('@open-wc/webpack-import-meta-loader'),
      })
    }
  }

Now get this issue:
client.js?06a0:102 TypeError: _memgraph_orb__WEBPACK_IMPORTED_MODULE_0__.Orb.Orb is not a constructor

Questions:
1. Is this an issue or am I doing something wrong?
2. What's the right way to solve this?

Thanks,
Ken

KenCarroll:
Hi @tonilastre created 'orbapp' to try to get orb working in a nuxtjs/vue project. But it wasn't the project where the above reported problem occurred. Was hoping to use 'orbapp' to try to replicate it...but have currently have a different problem (_memgraph_orb__WEBPACK_IMPORTED_MODULE_4__.Orb.Orb is not a constructor)

colek42:
I am having the same issue with Parcel 2.7.0.

It looks for the source files and can not find them.  If I clone the src files into nod modules, I get the issue @KenCarroll is having

```
[web] 🚨 Build failed.
[...]

PR #36, opened 2022-12-02
Fix web worker build with parcel

jondlm:
This fixed the issue #28 for me with Parcel. I believe the `module` is needed to instruct the bundler that it's an esmodule and should be bundled.

https://stackoverflow.com/questions/48045569/whats-the-difference-between-a-classic-and-module-web-worker

### DIFF
--- a/src/simulator/types/web-worker-simulator/simulator.ts
+++ b/src/simulator/types/web-worker-simulator/simulator.ts
@@ -16,6 +16,7 @@ export class WebWorkerSimulator extends Emitter<SimulatorEvents> implements ISim
         './process.worker',
         import.meta.url,
       ),
+      { type: 'module' }
     );
 
     this.worker.onmessage = ({ data }: MessageEvent<IWorkerOutputPayload>) => {


jondlm:
I was getting this error with parcel:
```
🚨 Build failed.
[Error: ENOENT: no such file or directory, open '/Users/jdlm/stripe/pay-server/frontend/graphy/node_modules/.pnpm/@memgraph+orb@0.1.4/node_modules/@memgraph/orb/src/simulator/types/web-worker-simulator/simulator.ts'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/jdlm/stripe/pay-server/frontend/graphy/node_modules/.pnpm/@memgraph+orb@0.1.4/node_modules/@memgraph/orb/src/simulator/types/web-worker-simulator/simulator.ts'
}
```

And this got it working.

I know this PR isn't quite up to contributing standards, but I figured I'd throw something up real quick so ya'll could take a look and tweak if you want it.

cizl:
Thanks a bunch for contributing!

Looking at [this browser compatibility list](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker#browser_compatibility), it seems that ES worker modules aren't supported by Firefox. I tried your branch and it seems to work fine for me. Could you please test your setup in Firefox to confirm that it works?

jondlm:
It works with Firefox (v107.0) on my Mac via Parcel. I believe that compatibility table doesn't apply in the Parcel case because the `{ type: 'module' }` acts has a hint for Parcel to bundle away the ES modules before sending it to the browser. Here's a snippet of what gets shipped to Firefox for `process.worker`:

```
// modules are defined as an array
// [ module function, map of requires ]
//
// map of requires is short require name -> numeric require
//
// anything defined in a previous bundle is accessed via the
// orig method which is the require for previous bundles

(function (modules, entry, mainEntry, parcelRequireName, globalName) {
  /* eslint-disable no-undef */
  var globalObject =
    typeof globalThis !== 'undefined'
      ? globalThis
      : typeof self !== 'undefined'
      ? self
      : typeof window !== 'undefined'
      ? window
      : typeof global !== 'undefined'
      ? global
      : {};
  /* eslint-enable no-undef */

  // Save the require from previous bundle to this closure if any
  var previousRequire =
    typeof globalObject[parcelRequireName] === 'function' &&
    globalObject[parcelRequireName];

...<snip>...
```

tonilastre (approved):

PR #37, opened 2022-12-02
New: Add simulator scenarios for manual testing

cizl:
This is only a helper PR for more clearer reviewing. All logic is in https://github.com/memgraph/orb/pull/38. The scenario matrix page is included here.
This is mostly testing and helper code, so it's not that significant, but @tonilastre feel free to add some pointers here. I've mostly hacked this as fast as I could.


tonilastre (approved):
This is cool!

Issue #18, opened 2022-09-21
Modify `orb.data.setup()` behavior to invalidate previous data


cizl:
Calling `orb.data.setup()` simply adds new nodes to existing ones, while the existing nodes stay fixed and unchanged.
This behavior would be expected for `orb.data.merge()` but not for `orb.data.setup()`

cizl:
https://user-images.githubusercontent.com/3769376/191435111-2b0e9f91-fb52-4ef8-ad00-1c5dec647883.mp4

PR #38, opened 2022-12-02
New: Refactor simulator (WIP)

cizl:
Rework simulator logic to better handle physics changes and data updates. Fixes https://github.com/memgraph/orb/issues/18
(WIP)

### DIFF
--- a/src/simulator/engine/d3-simulator-engine.ts
+++ b/src/simulator/engine/d3-simulator-engine.ts
@@ -72,6 +72,8 @@ export interface ID3SimulatorEngineSettingsPositioning {
 }
 
 export interface ID3SimulatorEngineSettings {
+  isSimulatingOnDataUpdate: boolean;
+  isSimulatingOnSettingsUpdate: boolean;
   isPhysicsEnabled: boolean;
   alpha: ID3SimulatorEngineSettingsAlpha;
   centering: ID3SimulatorEngineSettingsCentering | null;
@@ -89,12 +91,14 @@ export const getManyBodyMaxDistance = (linkDistance: number) => {
 };
 
 export const DEFAULT_SETTINGS: ID3SimulatorEngineSettings = {
+  isSimulatingOnDataUpdate: true,
+  isSimulatingOnSettingsUpdate: true,
   isPhysicsEnabled: false,
   alpha: {
     alpha: 1,
     alphaMin: 0.001,
     alphaDecay: 0.0228,
-    alphaTarget: 0.1,
+    alphaTarget: 0,
   },
   centering: {
     x: 0,
@@ -161,9 +165,11 @@ export type D3SimulatorEvents = {
 };
 
 export class D3SimulatorEngine extends Emitter<D3SimulatorEvents> {
-  protected readonly linkForce: ForceLink<ISimulationNode, SimulationLinkDatum<ISimulationNode>>;
-  protected readonly simulation: Simulation<ISimulationNode, undefined>;
-  protected readonly settings: ID3SimulatorEngineSettings;
+  protected linkForce!: ForceLink<ISimulationNode, SimulationLinkDatum<ISimulationNode>>;
+  protected simulation!: Simulation<ISimulationNode, undefined>;
+  // TODO(dlozic): Question: I removed readonly here since I'm using _initialSettings to reassign
+  // the settings in resetSettings(). Is this okay? Should I use Object.assign()?
+  protected settings!: ID3SimulatorEngineSettings;
 
   protected _edges: ISimulationEdge[] = [];
   protected _nodes: ISimulationNode[] = [];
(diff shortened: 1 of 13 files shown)


src/simulator/engine/d3-simulator-engine.ts
```diff
@@ -497,34 +557,152 @@ export class D3SimulatorEngine extends Emitter<D3SimulatorEvents> {
...
+   * Sticks all nodes thath have a defined position (x and y coordinates).
+   * This function should be called when the user initially sets up or merges some data.
+   * If the user provided nodes already have defined `x` **or** `y` properties, they are treated as _"sticky"_.
+   * Only the specified axis gets immobilized.
+   *
+   * @param {ID3SimulatorGraph} data Graph data.
+   * @return {ID3SimulatorGraph} Graph data with attached `{fx, sx}`, and/or `{fy, sy}` coordinates.
+   */
+  private _fixDefinedNodes(data: ID3SimulatorGraph): ID3SimulatorGraph {
+    // TODO(dlozic): Question: should this function be extracted or should i use `this.data` everywhere and remove inputs/outputs?
```
tonilastre, line 649: Well, here is how I see it: I am bit confused with the name, it says `fixDefinedNodes`. Ok, definition is, I assume if `x` or `y` is defined. Then it sets `fx` and `sx`. When I check the `fixNodes` (`fixNode`) method, it is not working like that, it is just setting `fx`. So, to recap, there is some inconsistency here - naming-wise.

tonilastre: Regarding the `data` input, as far as I can see it is used correctly (where `fixDefinedNodes` is used), meaning `data` is valid as an input here.

src/simulator/engine/d3-simulator-engine.ts
```diff
@@ -161,9 +165,11 @@ export type D3SimulatorEvents = {
 };
 
 export class D3SimulatorEngine extends Emitter<D3SimulatorEvents> {
-  protected readonly linkForce: ForceLink<ISimulationNode, SimulationLinkDatum<ISimulationNode>>;
-  protected readonly simulation: Simulation<ISimulationNode, undefined>;
-  protected readonly settings: ID3SimulatorEngineSettings;
+  protected linkForce!: ForceLink<ISimulationNode, SimulationLinkDatum<ISimulationNode>>;
+  protected simulation!: Simulation<ISimulationNode, undefined>;
+  // TODO(dlozic): Question: I removed readonly here since I'm using _initialSettings to reassign
+  // the settings in resetSettings(). Is this okay? Should I use Object.assign()?
```
tonilastre, line 171: Hmm it is a bit dirty, I think you can resolve both with this approach:

- always set `initialSettings`, currently initial settings are set only if there is constructor `settings`. You could have something like this:

```
// constructor
this.initialSettings = Object.assing(copyObject(DEFAULT_SETTINGS), settings ?? {});
```

You wan't to skip having `var: T | undefined` if it doesn't need to be undefined. By checking your code, having `initialSettings` as undefined doesn't make sense because even in `resetSettings` you are again using DEFAULT_SETTINGS instead of initialSettings. So, let's do all of that at once in the constructor. With that, your `initialSettings` will never be `undefined`.

- reset settings can be this:

```
resetSettings() {
  return copyObject(this._initialSettings);
}
```

- in constructor your settings can be then:

```
this.settings = resetSettings();
// or even
this.settings = copyObject(this._initialSettings);
```

What I want to say, whoever you want to reuse a function in the constructor, then make that function in a sense that it returns updated values, so you can do this in the constructor:

```
this.var = myFunc();
```

With that, you don't need to have `!` at the end of the name then.

Additionally, it is way better code, because if someone forgets to set a value in `myFunc()` (like now), because of the `variable!` definition, TS can't catch that.

src/simulator/engine/d3-simulator-engine.ts
```diff
@@ -172,33 +178,38 @@ export class D3SimulatorEngine extends Emitter<D3SimulatorEvents> {
...
+   * Applies the specified settings to the D3 simulator engine.
+   *
+   * @param {ID3SimulatorEngineSettingsUpdate} settings Partial D3 simulator engine settings (any property of settings)
+   */
   setSettings(settings: ID3SimulatorEngineSettingsUpdate) {
+    if (!this._initialSettings) {
+      this._initialSettings = Object.assign(copyObject(DEFAULT_SETTINGS), settings);
+    }
+
+    // TODO(dlozic): Question: (from line 166, 167) is this then necessary? Why not simple assign?
```
tonilastre, line 212: I think you are right. It should be double-checked.

src/views/orb-view.ts
```diff
@@ -212,26 +249,20 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
         }
       }
     }
+  };
 
-    this._isSimulating = true;
-    this._onSimulationEnd = onRendered;
-    this._startSimulation();
+  render(onRendered?: () => void) {
+    this._renderer.render(this._graph);
```
tonilastre, line 264: Btw what happens in this situation:

```
const orb = new OrbView();
// This will call the simulator in web-worker thread
orb.setup({ very large graph })

// What happens now? When will recenter happen, after the simulation, during the simulation or?
orb.render(() => {
  orb.recenter();
});
```

tonilastre (changes requested):
I like the approach where simulation happens on top of the data changes + sticky/fix support.

I've added several comments: there are a few bugs that I saw (e.g. calling [key] on undefined) that should be fixed, and all other comments are for discussion.

src/simulator/engine/d3-simulator-engine.ts
```diff
@@ -317,83 +346,114 @@ export class D3SimulatorEngine extends Emitter<D3SimulatorEvents> {
...
-    this.simulation.nodes();
-    this.linkForce.links();
+    this._updateSimulationData();
+  }
+
+  /**
+   * Resets the simulator engine by discarding all existing simulator data (nodes and edges),
+   * and keeping the current simulator engine settings.
+   */
+  private _resetSimulation() {
```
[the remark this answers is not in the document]
cizl: Yep, makes sense. Looking back, do you think this method should be public maybe?

src/simulator/engine/d3-simulator-engine.ts
cizl: If I have `_initialSettings` always defined (e.g. `DEFAULT_SETTINGS`) then they lose their purpose. I use this variable to store the initially defined settings. So when someone resets the settings, they don't default to `DEFAULT_SETTINGS`, but rather their initially defined settings. _(maybe this is not necessary, but I had a use case for this)_

Why? Because, in a worker context, we need to send the settings via a message - it can't be in the `OrbView` constructor. In `orb-view.ts` we can't provide settings during the creation of `d3-simulator-engine` because we can't send that object to the worker when the worker creates the `D3SimulatorEngine` in a separate thread.

```TypeScript
  const orb = new Orb.OrbView(container);
  orb.setSettings({
    simulation: {
    collision: {
      ...
     },
  });
```

So now, If they are always defined, I need to introduce another variable - `areSettingsInitialized` or something similar to know if the settings have been initialized. Maybe this is cleaner? What do you think?

I remember not being satisfied with this pattern but I implemented it this way because of the use case above.

cizl: Essentially, we don't have a way for the user to provide their own initial/default settings. The way I implemented it, the settings that he first assigns are treated as initial/default settings.

But now after writing this out, I guess he can simply keep a constant variable `const MY_DEFAULTS` and use `setSettings(MY_DEFAULTS)` instead of us trying to make `resetSettings` work.

src/views/orb-view.ts
cizl: https://user-images.githubusercontent.com/3769376/212104176-a83af1de-c54c-4277-82af-3288e60128a3.mp4


https://user-images.githubusercontent.com/3769376/212104189-13967fe7-bcad-4741-b60c-7269b4ce4275.mp4

src/simulator/engine/d3-simulator-engine.ts
cizl: > 

Yeah you're right. Should be `_stickDefinedNodes`.
Regarding naming, do the other methods make sense? (`fix` <-> `unfix`, `stick` <-> `unstick`)?

cizl: > Regarding the `data` input, as far as I can see it is used correctly (where `fixDefinedNodes` is used), meaning `data` is valid as an input here.

But should the method here go outside the class? Should it be a function and not a class method?

cizl: <rename and remove (release)>

src/simulator/engine/d3-simulator-engine.ts
cizl: <just check what happens if value is null>

Issue #19, opened 2022-09-23
Add canvas support for retina displays


tonilastre:

Fil:
thank you. It could be a parameter defaulting to devicePixelRatio, as in https://github.com/observablehq/stdlib/blob/main/src/dom/context2d.mjs#L2

tonilastre:
A parameter (to the Orb's renderer) seems like a good option. Thanks for the link!

tonilastre:
Will be released in v1.0.0.

PR #45, opened 2023-02-24
New: Add support for handling device pixel ratio

tonilastre:
The following PR adds support for automatic and manual handling of `devicePixelRatio`. More information about `devicePixelRatio` can be found here: https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio

There are two modes: Automatic and Manual.

*Automatic `devicePixelRatio` handler (default)*
The default behavior is the automatic handler of the `devicePixelRatio` value and changes (e.g. user drags a view from one display to another with a different DPR value). The automatic handler is turned on when the settings value `render.devicePixelRatio` equals to `null`.

*Manual `devicePixelRatio` handler*
If you want to manually handle the `devicePixelRatio` values, just set the settings value `render.devicePixelRatio` to a numeric DPR value. Orb will stop listening for DPR changes then.

This fixes issue #19.

### DIFF
--- a/src/renderer/canvas/canvas-renderer.ts
+++ b/src/renderer/canvas/canvas-renderer.ts
@@ -134,7 +134,7 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
     }
 
     // Change DPR from automatic to manual handling or change DPR value manually
-    if (!isNumber(previousDprValue) && isNumber(newDprValue) && newDprValue !== previousDprValue) {
+    if (!isNumber(previousDprValue) && isNumber(newDprValue)) {
       this._dprObserveUnsubscribe?.();
       this._resize();
     }
--- a/src/renderer/canvas/canvas-renderer.ts
+++ b/src/renderer/canvas/canvas-renderer.ts
@@ -18,8 +18,14 @@ import {
 import { throttle } from '../../utils/function.utils';
 import { getThrottleMsFromFPS } from '../../utils/math.utils';
 import { copyObject } from '../../utils/object.utils';
-import { appendCanvas, setupContainer } from '../../utils/html.utils';
+import {
+  appendCanvas,
+  IObserveDPRUnsubscribe,
+  observeDevicePixelRatioChanges,
+  setupContainer,
+} from '../../utils/html.utils';
 import { OrbError } from '../../exceptions';
+import { isNumber } from '../../utils/type.utils';
 
 const DEBUG = false;
 const DEBUG_RED = '#FF5733';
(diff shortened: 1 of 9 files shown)


src/renderer/canvas/canvas-renderer.ts
```diff
@@ -82,6 +132,17 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
         this._render(graph);
       }, getThrottleMsFromFPS(this._settings.fps));
     }
+
+    // Change DPR from automatic to manual handling or change DPR value manually
+    if (!isNumber(previousDprValue) && isNumber(newDprValue) && newDprValue !== previousDprValue) {
```
AlexIchenskiy, line 137: Nitpicking, but is `newDprValue !== previousDprValue` necessary here? It seems it's always true because `newDprValue` is a number while `previousDprValue` is not. On the other hand, `!isNumber(previousDprValue)` can be removed to allow users to change the DPR value more than once, but it does not seem like a real use case.

src/renderer/canvas/canvas-renderer.ts
```diff
@@ -195,6 +256,23 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
     }
   }
 
+  private _resize() {
+    const dpr = this._settings.devicePixelRatio || window.devicePixelRatio;
```
AlexIchenskiy, line 260: Nitpicking, but it would be nice to add a check for `window.devicePixelRatio` being null or undefined, just in case of older browsers that do not support this. I think using `|| 1` will do the job.

src/renderer/canvas/canvas-renderer.ts
tonilastre: Yep you are right, `newDrpValue !== previousDprValue` can be removed, but not the other check.

This is removing automatic observer for DPR. You can see on line 87 that it checks if settings doesn't have DPR, it will start automatic observer that checked for DPR from the browser.

So these lines 137 and 143 are for:
- 137: Stop the automatic observer - this can happen only if previousDPR was not a number, and the new value is defined
- 142: Start the observer again - this can happen only if previousDRP was a number, and the new value is null (this enables users to remove manual DPR handling).

Users can still change DPR manually which happens on line 127.

src/renderer/canvas/canvas-renderer.ts
tonilastre: Not nitpicking, but it's a valid bug. Great catch.

tonilastre, line 260:
```suggestion
    const dpr = this._settings.devicePixelRatio || window.devicePixelRatio || 1;
```

src/renderer/canvas/canvas-renderer.ts
tonilastre, line 137:
```suggestion
    if (!isNumber(previousDprValue) && isNumber(newDprValue)) {
```

AlexIchenskiy (approved):
Looks good to me now!

PR #46, opened 2023-04-03
New: Added support for context menu events

shashankshukla96:
This PR will add support for context menu events (mouse right click), the library exposes the 3 new events 
```
1. NODE_RIGHT_CLICK  :  Triggered when a node on the canvas is clicked
2. EDGE_RIGHT_CLICK : Triggered when an edge is clicked.
3. CANVAS_RIGHT_CLICK : Triggered when the canvas is clicked.
``` 

For example, to handle NODE_RIGHT_CLICK event we can use

```
orb.events.on(Orb.OrbEventType.NODE_RIGHT_CLICK, (data) => {
      data.event.preventDefault();
      console.log("Node Right Clicked", data);
})
```

### DIFF
--- a/examples/example-graph-events.html
+++ b/examples/example-graph-events.html
@@ -113,6 +113,22 @@
       console.log('Event: edge-click', event);
       output.innerHTML = `Edge with id <b>${event.edge.data.id}</b> clicked!`;
     });
+
+    orb.events.on(Orb.OrbEventType.NODE_RIGHT_CLICK, (data) => {
+      data.event.preventDefault();
+      console.log("Node Right Clicked", data);
+    })
+
+    orb.events.on(Orb.OrbEventType.EDGE_RIGHT_CLICK, (data) => {
+      data.event.preventDefault();
+      console.log("Edge Right Clicked", data);
+    })
+
+    orb.events.on(Orb.OrbEventType.CANVAS_RIGHT_CLICK, (data) => {
+      data.event.preventDefault();
+      console.log("Canvas Right Clicked", data);
+    })
+
   </script>
 </body>
 </html>
--- a/examples/example-graph-events.html
+++ b/examples/example-graph-events.html
@@ -116,17 +116,22 @@
 
     orb.events.on(Orb.OrbEventType.NODE_RIGHT_CLICK, (data) => {
       data.event.preventDefault();
-      console.log("Node Right Clicked", data);
+      console.log("Event: node-right-click", data);
+      output.innerHTML = `<b>${data.node.data.label}</b> right clicked!`;
     })
 
     orb.events.on(Orb.OrbEventType.EDGE_RIGHT_CLICK, (data) => {
       data.event.preventDefault();
-      console.log("Edge Right Clicked", data);
+      console.log("Event: edge-right-click", data);
+      output.innerHTML = `Edge with id <b>${data.edge.data.id}</b> right clicked!`;
     })
 
-    orb.events.on(Orb.OrbEventType.CANVAS_RIGHT_CLICK, (data) => {
+    orb.events.on(Orb.OrbEventType.MOUSE_RIGHT_CLICK, (data) => {
       data.event.preventDefault();
-      console.log("Canvas Right Clicked", data);
+      console.log("Event: mouse-right-click (canvas)", data);
+      if (!data.subject) {
+        output.innerHTML = `Canvas right clicked!`;
+      }
     })
 
   </script>
--- a/src/events.ts
+++ b/src/events.ts
@@ -26,7 +26,7 @@ export enum OrbEventType {
   NODE_DRAG_END = 'node-drag-end',
   NODE_RIGHT_CLICK = 'node-right-click',
   EDGE_RIGHT_CLICK = 'edge-right-click',
-  CANVAS_RIGHT_CLICK = 'canvas-right-click',
+  MOUSE_RIGHT_CLICK = 'mouse-right-click',
 }
 
 export interface IOrbEventDuration {
@@ -88,5 +88,5 @@ export class OrbEmitter<N extends INodeBase, E extends IEdgeBase> extends Emitte
   [OrbEventType.NODE_DRAG_END]: IOrbEventMouseNodeEvent<N, E> & IOrbEventMouseMoveEvent;
   [OrbEventType.NODE_RIGHT_CLICK]: IOrbEventMouseNodeEvent<N, E> & IOrbEventMouseClickEvent;
   [OrbEventType.EDGE_RIGHT_CLICK]: IOrbEventMouseEdgeEvent<N, E> & IOrbEventMouseClickEvent;
-  [OrbEventType.CANVAS_RIGHT_CLICK]: IOrbEventMouseClickEvent;
+  [OrbEventType.MOUSE_RIGHT_CLICK]:  IOrbEventMouseEvent<N, E> & IOrbEventMouseClickEvent;
 }> {}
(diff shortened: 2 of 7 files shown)


src/events.ts
```diff
@@ -83,4 +86,7 @@ export class OrbEmitter<N extends INodeBase, E extends IEdgeBase> extends Emitte
   [OrbEventType.NODE_DRAG_START]: IOrbEventMouseNodeEvent<N, E> & IOrbEventMouseMoveEvent;
   [OrbEventType.NODE_DRAG]: IOrbEventMouseNodeEvent<N, E> & IOrbEventMouseMoveEvent;
   [OrbEventType.NODE_DRAG_END]: IOrbEventMouseNodeEvent<N, E> & IOrbEventMouseMoveEvent;
+  [OrbEventType.NODE_RIGHT_CLICK]: IOrbEventMouseNodeEvent<N, E> & IOrbEventMouseClickEvent;
+  [OrbEventType.EDGE_RIGHT_CLICK]: IOrbEventMouseEdgeEvent<N, E> & IOrbEventMouseClickEvent;
+  [OrbEventType.CANVAS_RIGHT_CLICK]: IOrbEventMouseClickEvent;
```
tonilastre, line 91:
```suggestion
  [OrbEventType.MOUSE_RIGHT_CLICK]: IOrbEventMouseEvent<N, E> & IOrbEventMouseClickEvent;
```
I would suggest having a bit more generic event such as `mouse-right-click` that can handle any of the following three scenarios:

* Node is right-clicked, `subject` will be `INode`
* Edge is right-clicked, `subject` will be `IEdge`
* Canvas is right-clicked, so no node and no edge, `subject` will be `undefined`.

This is how `MOUSE_CLICK` works with `NODE_CLICK` and `EDGE_CLICK`. You can listen to `MOUSE_CLICK` and handle all three cases there if you want. This also enables someone to make a custom radius selection on top of the event position.

src/views/default-view.ts
```diff
@@ -385,6 +386,49 @@ export class DefaultView<N extends INodeBase, E extends IEdgeBase> implements IO
     }
   };
 
+  mouseRightClicked = (event: PointerEvent) => {
+    // event.preventDefault();
```
tonilastre, line 390:
```suggestion
```

src/views/default-view.ts
```diff
@@ -385,6 +386,49 @@ export class DefaultView<N extends INodeBase, E extends IEdgeBase> implements IO
...
+          localPoint: simulationPoint,
+          globalPoint: mousePoint,
+        });
+      }
+
+      
+
+      // if (response.isStateChanged || response.changedSubject) {
+      //   this._renderer.render(this._graph);
+      // }
```
tonilastre, line 428:
```suggestion
      if (response.isStateChanged || response.changedSubject) {
        this._renderer.render(this._graph);
      }
```

Is there a particular reason why this is uncommented? The default click strategy you added selects a node or an edge which means its style will change and that will require a new render. That is the point of these lines.

src/views/default-view.ts
```diff
@@ -385,6 +386,49 @@ export class DefaultView<N extends INodeBase, E extends IEdgeBase> implements IO
...
+            globalPoint: mousePoint,
+          });
+        }
+      } else {
+        this._events.emit(OrbEventType.CANVAS_RIGHT_CLICK, {
+          event,
+          localPoint: simulationPoint,
+          globalPoint: mousePoint,
+        });
+      }
```
tonilastre, line 422:
```suggestion
      }
      
      this._events.emit(OrbEventType.MOUSE_RIGHT_CLICK, {
        subject,
        event,
        localPoint: simulationPoint,
        globalPoint: mousePoint,
      });
```

Just as my comment before regarding the more generic approach with the mouse click. In this case, the subject will be `undefined` where someone using the API can easily know if the mouse right click was on canvas or on top of the node or edge, e.g.

```typescript
orb.events.on(Orb.OrbEventType.MOUSE_RIGHT_CLICK, (event) => {
  if (!event.subject) {
    // This is a canvas click
  }
});
```

src/views/default-view.ts
```diff
@@ -385,6 +386,49 @@ export class DefaultView<N extends INodeBase, E extends IEdgeBase> implements IO
...
 
+  mouseRightClicked = (event: PointerEvent) => {
+    // event.preventDefault();
+    const mousePoint = this.getCanvasMousePosition(event);
+    const simulationPoint = this._renderer.getSimulationPosition(mousePoint);
+
+    if (this._strategy.onMouseRightClick) {
+      const response = this._strategy.onMouseRightClick(this._graph, simulationPoint);
+      const subject = response.changedSubject;
+
```
tonilastre, line 397:
```suggestion
```

examples/example-graph-events.html
```diff
@@ -113,6 +113,22 @@ <h1>Example 4 - Events</h1>
       console.log('Event: edge-click', event);
       output.innerHTML = `Edge with id <b>${event.edge.data.id}</b> clicked!`;
     });
+
+    orb.events.on(Orb.OrbEventType.NODE_RIGHT_CLICK, (data) => {
+      data.event.preventDefault();
+      console.log("Node Right Clicked", data);
```
tonilastre, line 119:
```suggestion
      console.log("Event: node-right-click", data);
      output.innerHTML = `<b>${data.node.data.label}</b> right clicked!`;
```

I would suggest adding these so the user can see the event message on the UI element.

examples/example-graph-events.html
```diff
@@ -113,6 +113,22 @@ <h1>Example 4 - Events</h1>
...
+    })
+
+    orb.events.on(Orb.OrbEventType.EDGE_RIGHT_CLICK, (data) => {
+      data.event.preventDefault();
+      console.log("Edge Right Clicked", data);
```
tonilastre, line 124:
```suggestion
      console.log("Event: edge-right-click", data);
      output.innerHTML = `Edge with id <b>${data.edge.data.id}</b> right clicked!`
```

I would suggest adding these so the user can see the event message on the UI element.

examples/example-graph-events.html
```diff
@@ -113,6 +113,22 @@ <h1>Example 4 - Events</h1>
...
+    })
+
+    orb.events.on(Orb.OrbEventType.CANVAS_RIGHT_CLICK, (data) => {
+      data.event.preventDefault();
+      console.log("Canvas Right Clicked", data);
```
tonilastre, line 129:
```suggestion
      console.log("Event: mouse-right-click (canvas)", data);
      if (!data.subject) {
        output.innerHTML = `Canvas right clicked!`
      }
```

I would suggest adding these so the user can see the event message on the UI element.

src/views/map-view.ts
```diff
@@ -252,10 +252,46 @@ export class MapView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+            localPoint: point,
+            globalPoint: containerPoint,
+          });
+        }
+
+        
+
+        if (response.isStateChanged) {
+          this._renderer.render(this._graph);
+        }
```
tonilastre, line 292:
```suggestion
        }
        
        this._events.emit(OrbEventType.MOUSE_RIGHT_CLICK, {
          subject,
          event: event.originalEvent,
          localPoint: point,
          globalPoint: containerPoint,
        });

        if (response.isStateChanged || response.changedSubject) {
          this._renderer.render(this._graph);
        }
```

Same comment as for the default view, it would be great to have a higher abstraction for the general right click.

tonilastre (changes requested):
Thanks a lot for your contribution! I've added a few comments, more about alignment with the existing mouse (left) events.

tonilastre (approved):

PR #50, opened 2023-04-24
New: Added support for double click events

shashankshukla96:
This PR will add support for double click events, the library exposes the 3 new events 
```
1. NODE_DOUBLE_CLICK :  Triggered when a node on the canvas is double clicked
2. EDGE_DOUBLE_CLICK : Triggered when an edge is double clicked.
3. MOUSE_DOUBLE_CLICK: Triggered when the canvas is double clicked.
``` 

For example, to handle NODE_DOUBLE_CLICK event we can use

```
orb.events.on(Orb.OrbEventType.NODE_DOUBLE_CLICK , (data) => {
      console.log("Node Double Clicked", data);
})
```

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -33,7 +33,11 @@ export enum OrbEventType {
   NODE_DRAG_END = 'node-drag-end',
   NODE_RIGHT_CLICK = 'node-right-click',
   EDGE_RIGHT_CLICK = 'edge-right-click',
-  CANVAS_RIGHT_CLICK = 'canvas-right-click',
+  MOUSE_RIGHT_CLICK = 'mouse-right-click',
+  // DBL click events
+  NODE_DOUBLE_CLICK = 'node-double-click',
+  EDGE_DOUBLE_CLICK = 'edge-double-click',
+  MOUSE_DOUBLE_CLICK = 'mouse-double-click',  
 }
 ```
 
--- a/src/views/map-view.ts
+++ b/src/views/map-view.ts
@@ -252,7 +252,7 @@ export class MapView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
 
     // Leaflet doesn't have a valid type definition for click event
     // @ts-ignore
-    leaflet.on('click contextmenu', (event: ILeafletEvent<PointerEvent>) => {
+    leaflet.on('click contextmenu dblclick', (event: ILeafletEvent<PointerEvent>) => {
       const point: IPosition = { x: event.layerPoint.x, y: event.layerPoint.y };
       const containerPoint: IPosition = { x: event.containerPoint.x, y: event.containerPoint.y };
 
@@ -290,8 +290,7 @@ export class MapView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
           this._renderer.render(this._graph);
         }
         return;
-      }
-      if (event.type === 'click' && this._strategy.onMouseClick) {
+      } else if (event.type === 'click' && this._strategy.onMouseClick) {
         const response = this._strategy.onMouseClick(this._graph, point);
         const subject = response.changedSubject;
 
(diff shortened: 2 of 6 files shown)


docs/events.md
```diff
@@ -33,7 +33,11 @@ export enum OrbEventType {
   NODE_DRAG_END = 'node-drag-end',
   NODE_RIGHT_CLICK = 'node-right-click',
   EDGE_RIGHT_CLICK = 'edge-right-click',
-  CANVAS_RIGHT_CLICK = 'canvas-right-click',
+  MOUSE_RIGHT_CLICK = 'mouse-right-click',
```
tonilastre, line 36: Great catch here!

src/views/map-view.ts
```diff
@@ -290,8 +290,7 @@ export class MapView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
           this._renderer.render(this._graph);
         }
         return;
-      }
-      if (event.type === 'click' && this._strategy.onMouseClick) {
+      } else if (event.type === 'click' && this._strategy.onMouseClick) {
```
tonilastre, line 293: You have `return` in the line before so you don't need `else if`. Either remove the `return` or keep `if + return` for all three cases.

src/views/map-view.ts
```diff
@@ -321,6 +320,39 @@ export class MapView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
           globalPoint: containerPoint,
         });
 
+        if (response.isStateChanged || response.changedSubject) {
+          this._renderer.render(this._graph);
+        }
+      } else if (event.type === 'dblclick' && this._strategy.onMouseDoubleClick) {
```
tonilastre, line 326: There is a problem with the map view because double click has two actions: One does the event of DOUBLE_CLICK, but the other one does the leaflet zoom (because of the double click).

If there is a strategy and the subject is either `node` or an `edge`, it would be great to prevent the default leaflet zoom operation on double click. If there is a double click on canvas (no subject), the leaflet should do the default action on double click which is a zoom.
You can play with this in the "Map view" example.

tonilastre (changes requested):
Btw, please run the `npm run lint` (and `npm run lint:fix`) to fix lint issues. I've added automatic lint in the baseline branch that will be used for your PR: #51

src/views/map-view.ts
```diff
@@ -321,6 +321,42 @@ export class MapView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+          this._events.emit(OrbEventType.MOUSE_DOUBLE_CLICK, {
+            subject,
+            event: event.originalEvent,
+            localPoint: point,
+            globalPoint: containerPoint,
+          });
+          // zoom in on double click
+          const zoom = event.target._zoom + 1;
+          event.target.setZoomAround(event.layerPoint, zoom);
+        }
```
tonilastre, line 358:
```suggestion
        }
        
        this._events.emit(OrbEventType.MOUSE_DOUBLE_CLICK, {
          subject,
          event: event.originalEvent,
          localPoint: point,
          globalPoint: containerPoint,
        });
        
        // zoom in on double click if no subject underneath
        if (!subject) {
          const zoom = event.target._zoom + 1;
          event.target.setZoomAround(event.layerPoint, zoom);
        }
```

The current code won't send MOUSE_DOUBLE_CLICK events (even if the subject exists) which the user might listen for. All other general mouse events send MOUSE_CLICK events for both cases if the subject is defined or not defined, so we should keep the same behavior here too.

tonilastre (changes requested):
Great fix. There is one issue left around the consistency of events, so please check it out!

src/views/map-view.ts
shashankshukla96: yes got it. Thanks !

tonilastre (approved):
Looks great. Please run `npm run lint:fix` once more and commit, there are 2 standing issues. Btw, from the next version forward, linting will be in the pre-commit hook.

PR #53, opened 2023-05-01
New: Add zIndex style property

tonilastre:
Additional `zIndex` property on nodes and edges that act similar to CSS z-index. If set on any of the nodes, the zIndex property value will be used as a stack order when rendering nodes/edges.

An example "Custom styled graph" has been updated where the first node has the highest zIndex value.

Screenshot 2023-05-01 at 12 25 16

### DIFF
--- a/src/renderer/canvas/canvas-renderer.ts
+++ b/src/renderer/canvas/canvas-renderer.ts
@@ -132,8 +132,8 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
       this._context.fillRect(0, 0, this.width, this.height);
     }
 
-    this.drawObjects(graph.getEdges());
-    this.drawObjects(graph.getNodes());
+    this.drawObjects(sortObjectsByZIndex(graph.getEdges()));
+    this.drawObjects(sortObjectsByZIndex(graph.getNodes()));
 
     this._context.restore();
     this.emit(RenderEventType.RENDER_END, { durationMs: Date.now() - renderStartedAt });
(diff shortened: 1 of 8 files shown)


src/renderer/canvas/canvas-renderer.ts
```diff
@@ -132,8 +132,8 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
       this._context.fillRect(0, 0, this.width, this.height);
     }
 
-    this.drawObjects(graph.getEdges());
-    this.drawObjects(graph.getNodes());
+    this.drawObjects(sortObjectsByZIndex(graph.getEdges()));
+    this.drawObjects(sortObjectsByZIndex(graph.getNodes()));
```
cizl, line 136: Is this expensive? Can we keep the ordering contained inside the data structure for the nodes and edges?
Because this function gets called for every render cycle and if there are a bunch of elements this seems costly.
It seems to me that it would be wider to call a reorder function when we update the graph, rather than on every render cycle.

cizl: I'm fine with leaving it like it is for now, but definitely write it down and take this into consideration when refactoring for v1.0.

Allowing the user to directly access the props is proving more and more to be tricky to handle.

cizl (approved):

src/renderer/canvas/canvas-renderer.ts
cizl: So for now I'd just plop in a TODO here :)

tonilastre: I've added `EntityModel` and changed the code in such a way that ordering will happen on one of the following events:
* `data.setup`
* `data.merge`
* `data.remove`
* `setDefaultStyle`

And when we add reactivity, we can then refactor all of these to be more reactive on each node/edge style changes.

cizl (approved):
Amazing!

I think the entity stuff is a great improvement.

PR #55, opened 2023-05-03
New: Add canvas background color

tonilastre:
The following PR adds additional property (`render.backgroundColor`) that can be set via view settings. It supports hex colors and `Color` objects, including transparent colors (RGBA, HSLA, ...).

To change the background color, call the following code:

```typescript
orb.view.setSettings({
  render: {
    backgroundColor: '#DDDDDD',
  },
});

// reset the background color with
orb.view.setSettings({
  render: {
    backgroundColor: null,
  },
});
```

It works for both the default view and the map view. Map view is interesting when using colors with alpha (transparency channel):

Screenshot 2023-05-03 at 18 47 19
Screenshot 2023-05-03 at 18 46 22

### DIFF
--- a/src/renderer/canvas/canvas-renderer.ts
+++ b/src/renderer/canvas/canvas-renderer.ts
@@ -98,6 +98,10 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
 
     // Clear drawing.
     this._context.clearRect(0, 0, this.width, this.height);
+    if (this._settings.backgroundColor) {
+      this._context.fillStyle = this._settings.backgroundColor.toString();
+      this._context.fillRect(0, 0, this.width, this.height);
+    }
     this._context.save();
 
     if (DEBUG) {
--- a/docs/view-default.md
+++ b/docs/view-default.md
@@ -79,6 +79,7 @@ interface IDefaultViewSettings {
     shadowOnEventIsEnabled: boolean;
     contextAlphaOnEvent: number;
     contextAlphaOnEventIsEnabled: boolean;
+    backgroundColor: Color | string | null;
   };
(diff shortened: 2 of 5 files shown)


src/renderer/canvas/canvas-renderer.ts
```diff
@@ -98,6 +98,10 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
 
     // Clear drawing.
     this._context.clearRect(0, 0, this.width, this.height);
+    if (this._settings.backgroundColor) {
+      this._context.fillStyle = this._settings.backgroundColor.toString();
```
cizl, line 102: Since the `backgroundColor` can be a string, do we validate anywhere if it's a valid color hex?

Or do we leave this to the HTML5 canvas API? 

I tried this on the canvas and nothing happens if an unexpected value is input. Should we maybe warn the user somewhere?

tonilastre: Good note!

I've also tested and because it is not doing anything "wrong" on bad input, I left it as is because that is also how `canvas` works, meaning if it has something unhandled, it won't do it. The top-level usage, in our case the user of orb API, should handle the errors and validations of this matter. Same as how `canvas` is not breaking that we are using. Like a propagation. I would also skip warnings too in this case too because we are not doing any fallback action (like we do in the web-workers case).

What do you think?

cizl: Yep, that makes sense.
But still, the canvas behavior itself is a bit strange. Usually, successful actions are silent.

tonilastre: Agree :)

Issue #60, opened 2023-05-29
Request: Method to retrieve all selected nodes and edges


parmar-abhinav:
Request for the introduction of a new method in the library that allows users to retrieve all selected nodes and edges within the graph. Currently, there doesn't appear to be a straightforward way to access this information, which limits the flexibility and functionality of the library.

PR #61, opened 2023-05-29
New: Added support to get selected/hovered nodes and edges

parmar-abhinav:
Issue: #60 

**Description:**

This pull request addresses the issue raised regarding the lack of a straightforward way to retrieve selected/hovered nodes and edges in the library. To enhance the flexibility and functionality of the library, this PR introduces four new methods, `getSelectedNodes`, `getSelectedEdges`, `getHoveredNodes`, `getHoveredEdges`. Which enable users to retrieve all selected/hovered elements within the graph.

**Changes Made:**

- Implemented the `getSelectedNodes` method to retrieve all selected nodes.
- Implemented the `getSelectedEdges` method to retrieve all selected edges.
- Implemented the `getHoveredNodes` method to retrieve all hovered nodes.
- Implemented the `getHoveredEdges` method to retrieve all hovered edges.

**Example Usage:**

```typescript
// Retrieve selected nodes
selectedNodes = orb.data.getSelectedNodes()

// Retrieve selected edges
selectedEdges = orb.data.getSelectedEdges()

// Retrieve hovered nodes
hoveredNodes = orb.data.getHoveredNodes()

// Retrieve selected edges
hoveredEdges = orb.data.getHoveredEdges()

### DIFF
--- a/src/models/graph.ts
+++ b/src/models/graph.ts
@@ -24,6 +24,8 @@ export interface IGraph<N extends INodeBase, E extends IEdgeBase> {
   getEdgeById(id: any): IEdge<N, E> | undefined;
   getSelectedNodes(): INode<N, E>[];
   getSelectedEdges(): IEdge<N, E>[];
+  getHoveredNodes(): INode<N, E>[];
+  getHoveredEdges(): IEdge<N, E>[];
   getNodePositions(): INodePosition[];
   setNodePositions(positions: INodePosition[]): void;
   getEdgePositions(): IEdgePosition[];
@@ -136,6 +138,24 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N
    */
   getSelectedEdges(): IEdge<N, E>[] {
     return this.getEdges((edge) => edge.isSelected());
+  }
+
+  /**
+   * Returns a list of hovered nodes.
+   *
+   * @return {INode[]} List of hovered nodes
+   */
+  getHoveredNodes(): INode<N, E>[] {
+    return this.getNodes((node) => node.isHovered());
+  }
+
+  /**
+   * Returns a list of hovered edges.
+   *
+   * @return {IEdge[]} List of hovered edges
+   */
+  getHoveredEdges(): IEdge<N, E>[] {
+    return this.getEdges((edge) => edge.isHovered());
   }
 
   /**
--- a/src/models/graph.ts
+++ b/src/models/graph.ts
@@ -22,6 +22,8 @@ export interface IGraph<N extends INodeBase, E extends IEdgeBase> {
   getEdgeCount(): number;
   getNodeById(id: any): INode<N, E> | undefined;
   getEdgeById(id: any): IEdge<N, E> | undefined;
+  getSelectedNodes(): INode<N, E>[];
+  getSelectedEdges(): IEdge<N, E>[];
   getNodePositions(): INodePosition[];
   setNodePositions(positions: INodePosition[]): void;
   getEdgePositions(): IEdgePosition[];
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by tonilastre** [comment]
Thanks for the contribution! :D 

Btw, before checking this, was this working for you?

```typescript
const selectedNodes = orb.data.getNodes((node) => node.isSelected());
const selectedEdges = orb.data.getEdges((edge) => edge.isSelected());
```

The whole idea behind this optional `filterBy` parameter in the `getNodes/getEdges` was to enable a generic approach to get nodes/edges on a custom filter function. You as a lib user can also change the State enum and have your custom states where functions such as `getSelectedNodes` and `getSelectedEdges` might become obsolete because they do not match the states.
E.g. you might have custom states and custom click/hover strategies without any mention of "selected" or "hovered".

parmar-abhinav:
Thank you for your response! 

Yes below suggestion is working for me
```
const selectedNodes = orb.data.getNodes((node) => node.isSelected());
const selectedEdges = orb.data.getEdges((edge) => edge.isSelected());
```

I understand your point about the flexibility provided by the `filterBy` parameter in the `getNodes` and `getEdges` methods, which allows users to define custom filter functions.

However, I decided to introduce the `getSelectedNodes` and `getSelectedEdges` methods for a few reasons. First, selecting nodes is a common requirement in visualization libraries, and having a dedicated method to retrieve selected nodes can provide a more intuitive and convenient approach for users. Additionally, the existing `isSelected` method already indicates whether any nodes or edges are selected or not, so it made sense to complement it with methods specifically designed to return selected nodes and edges.

While it's true that users can modify the `State` enum and define their own custom states, it's also valuable to have predefined methods that align with common use cases like selection. These methods can save users time and effort by providing a straightforward solution without the need for custom filter functions.

Please let me know if you have any further questions or concerns.

tonilastre:
@parmar-abhinav Yep, agree with you. it would be also great to add `getHoveredNodes` and `getHoveredEdges` too - what do you think?

src/models/graph.ts
```diff
@@ -118,6 +120,28 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N
...
 
+  /**
+   * Returns a list of selected nodes.
+   *
+   * @return {INode[]} List of selected nodes
+   */
+  getSelectedNodes(): INode<N, E>[] {
+    const nodes = this.getNodes();
+    const selectedNodes = nodes.filter((node) => node.isSelected());
+    return selectedNodes;
```
tonilastre, line 131: I think it would be valuable to use the inner model for filtering. What I mean by this is to change this to

```typescript
return this.getNodes((node) => node.isSelected());
```

The inner model (`nodes`) keeps the ordering + it might have some other optimizations in the future. 

And what I commented in the main thread, it might be useful to have `getHoveredNodes` and `getHoveredEdges` - for all of these you can use the `getNodes` and `getEdges` with custom filter.

tonilastre (changes requested):

parmar-abhinav:
> @parmar-abhinav Yep, agree with you. it would be also great to add `getHoveredNodes` and `getHoveredEdges` too - what do you think?

Yes, I agree with your suggestion. Should I add `getHoveredNodes` and `getHoveredEdges` method as a part of this PR ?

tonilastre:
> > @parmar-abhinav Yep, agree with you. it would be also great to add `getHoveredNodes` and `getHoveredEdges` too - what do you think?
> 
> Yes, I agree with your suggestion. Should I add `getHoveredNodes` and `getHoveredEdges` method as a part of this PR ?

Yep, I think that makes sense. You can also update the title of the PR to match all changes. Thanks

parmar-abhinav:
@tonilastre Updated the PR as per suggestions, Pls check.

tonilastre (approved):

PR #63, opened 2023-06-05
New: Add support for label styling options

shashankshukla96:
This PR will add support to add label background styling options.

This will add style properties - 

 - fontBackgroundBorderWidth
 - fontBackgroundBorderColor
 - fontBackgroundBorderRadius
 - fontBackgroundPadding

### DIFF
--- a/src/renderer/canvas/label.ts
+++ b/src/renderer/canvas/label.ts
@@ -14,6 +14,10 @@ export enum LabelTextBaseline {
 
 export interface ILabelProperties {
   fontBackgroundColor: Color | string;
+  fontBackgroundBorderWidth: number;
+  fontBackgroundBorderColor: Color | string;
+  fontBackgroundBorderRadius: number;
+  fontBackgroundPadding: number;
   fontColor: Color | string;
   fontFamily: string;
   fontSize: number;
@@ -75,19 +79,53 @@ const drawTextBackground = (context: CanvasRenderingContext2D, label: Label) => 
 
   context.fillStyle = label.properties.fontBackgroundColor.toString();
   const margin = label.fontSize * FONT_BACKGROUND_MARGIN;
-  const height = label.fontSize + 2 * margin;
+
+  let height = label.fontSize + 2 * margin;
   const lineHeight = label.fontSize * FONT_LINE_SPACING;
+
   const baselineHeight = label.textBaseline === LabelTextBaseline.MIDDLE ? label.fontSize / 2 : 0;
 
   for (let i = 0; i < label.textLines.length; i++) {
     const line = label.textLines[i];
-    const width = context.measureText(line).width + 2 * margin;
-    context.fillRect(
-      label.position.x - width / 2,
-      label.position.y - baselineHeight - margin + i * lineHeight,
-      width,
-      height,
-    );
+
+    let width = context.measureText(line).width + 2 * margin;
+    let x = label.position.x - width / 2;
+    const y = label.position.y - baselineHeight - margin + i * lineHeight;
+
+    if (label.properties.fontBackgroundPadding) {
+      width += label.properties.fontBackgroundPadding * 2;
+      height += label.properties.fontBackgroundPadding * 2;
+      x -= label.properties.fontBackgroundPadding;
+    }
+
+    if (label.properties.fontBackgroundBorderRadius) {
+      const borderRadius = Math.min(label.properties.fontBackgroundBorderRadius, width / 2, height / 2);
+
+      context.beginPath();
+      context.lineWidth = label.properties.fontBackgroundBorderWidth ?? 0;
+
+      context.moveTo(x + borderRadius, y);
+      context.lineTo(x + width - borderRadius, y);
+      context.quadraticCurveTo(x + width, y, x + width, y + borderRadius);
+      context.lineTo(x + width, y + height - borderRadius);
+      context.quadraticCurveTo(x + width, y + height, x + width - borderRadius, y + height);
+      context.lineTo(x + borderRadius, y + height);
+      context.quadraticCurveTo(x, y + height, x, y + height - borderRadius);
+      context.lineTo(x, y + borderRadius);
+      context.quadraticCurveTo(x, y, x + borderRadius, y);
+      context.closePath();
+      context.fillStyle = (label.properties.fontBackgroundColor ?? DEFAULT_FONT_COLOR).toString();
+      context.strokeStyle = (label.properties.fontBackgroundBorderColor ?? DEFAULT_FONT_COLOR).toString();
+      context.fill();
+      context.stroke();
+    } else if (label.properties.fontBackgroundBorderWidth && !label.properties.fontBackgroundBorderRadius) {
+      context.lineWidth = label.properties.fontBackgroundBorderWidth;
+      context.strokeStyle = (label.properties.fontBackgroundBorderColor ?? DEFAULT_FONT_COLOR).toString();
+
+      context.strokeRect(x, y, width, height);
+    } else {
+      context.fillRect(x, y, width, height);
+    }
   }
 };
 
@@ -102,9 +140,16 @@ const drawText = (context: CanvasRenderingContext2D, label: Label) => {
   context.textAlign = 'center';
   const lineHeight = label.fontSize * FONT_LINE_SPACING;
 
+  const x = label.position.x;
+  let y = label.position.y;
+
+  if (label.properties.fontBackgroundPadding) {
+    y += label.properties.fontBackgroundPadding;
+  }
+
   for (let i = 0; i < label.textLines.length; i++) {
     const line = label.textLines[i];
-    context.fillText(line, label.position.x, label.position.y + i * lineHeight);
+    context.fillText(line, x, y + i * lineHeight);
   }
 };
 
--- a/docs/styles.md
+++ b/docs/styles.md
@@ -33,6 +33,10 @@ the following properties:
 | `colorHover`          | Color &#124; string | Node background color on mouse hover event. If not defined `color` is used. |
 | `colorSelected`       | Color &#124; string | Node background color on mouse click event. If not defined `color` is used. |
 | `fontBackgroundColor` | Color &#124; string | Node text (label) background color.             |
+| `fontBackgroundBorderWidth` | number        | Node text (label) background border width.      |
+| `fontBackgroundBorderColor` | Color &#124; string | Node text (label) background border color.   |
+| `fontBackgroundBorderRadius` | number        | Node text (label) background border radius.     |
+| `fontBackgroundPadding` | number            | Node text (label) background padding.           |
 | `fontColor`           | Color &#124; string | Node text (label) font color. The default is `#000000`. |
 | `fontFamily`          | string              | Node text (label) font family. The default is `"Roboto, sans-serif"`. |
 | `fontSize`            | number              | Node text (label) font size. The default is `4`.    |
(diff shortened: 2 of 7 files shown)


src/renderer/canvas/label.ts
```diff
@@ -75,19 +79,53 @@ const drawTextBackground = (context: CanvasRenderingContext2D, label: Label) =>
...
-      height,
-    );
+
+    let width = context.measureText(line).width + 2 * margin;
+    let x = label.position.x - width / 2;
+    const y = label.position.y - baselineHeight - margin + i * lineHeight;
+
+    if (label.properties.fontBackgroundPadding) {
+      width += label.properties.fontBackgroundPadding * 2;
+      height += label.properties.fontBackgroundPadding * 2;
```
tonilastre, line 97: I feel like the `margin` used internally here is the same as this `padding` so it would be great to merge those and have some value as default. Currently, the default margin is `1.2`.

src/renderer/canvas/label.ts
```diff
@@ -75,19 +79,53 @@ const drawTextBackground = (context: CanvasRenderingContext2D, label: Label) =>
...
+      context.moveTo(x + borderRadius, y);
+      context.lineTo(x + width - borderRadius, y);
+      context.quadraticCurveTo(x + width, y, x + width, y + borderRadius);
+      context.lineTo(x + width, y + height - borderRadius);
+      context.quadraticCurveTo(x + width, y + height, x + width - borderRadius, y + height);
+      context.lineTo(x + borderRadius, y + height);
+      context.quadraticCurveTo(x, y + height, x, y + height - borderRadius);
+      context.lineTo(x, y + borderRadius);
+      context.quadraticCurveTo(x, y, x + borderRadius, y);
+      context.closePath();
```
tonilastre, line 116: I think you can use `drawRoundRect` from `shapes` here.

src/renderer/canvas/label.ts
```diff
@@ -75,19 +79,53 @@ const drawTextBackground = (context: CanvasRenderingContext2D, label: Label) =>
...
+      context.fillStyle = (label.properties.fontBackgroundColor ?? DEFAULT_FONT_COLOR).toString();
```
tonilastre, line 117: Are you sure you want to use `DEFAULT_FONT_COLOR` here? If default font color is black, and the background is black too, then by default, the text won't be visible.

src/renderer/canvas/label.ts
```diff
@@ -75,19 +79,53 @@ const drawTextBackground = (context: CanvasRenderingContext2D, label: Label) =>
...
+      context.strokeStyle = (label.properties.fontBackgroundBorderColor ?? DEFAULT_FONT_COLOR).toString();
+      context.fill();
+      context.stroke();
+    } else if (label.properties.fontBackgroundBorderWidth && !label.properties.fontBackgroundBorderRadius) {
```
tonilastre, line 121: You already have `if(label.properties.fontBackgroundBorderRadius)` on line 101, to the check for (!label.properties.fontBackgroundBorderRadius` is not needed here, it is irrelevant.

tonilastre: It is hard to figure out what will be the shape drawn from these two ifs and else so I propose splitting it so it is much easier to understand what final shape will be drawn, e.g.

```
// setup and preps

const isBackgroundShapeDrawn = ....
if (!isBackgroundShapeDrawn) {
  continue;
}

// I think this function will also handle radius = 0 to draw proper rect
drawRoundRect(...);

if (backgroundColor) {
  setFill = 
  fill();
}

If (borderWidth) {
  setStrokeStyle = 
  stroke();
}
```

tonilastre (changes requested):
There is one additional bug. When using multi-line labels, there is no background. Just change the line 36 in the `example-custom-styled-graph` to be:

```
      { id: 1, label: 'Node B - A Special Node\nNew Line' },
```

And you will get this:

Screenshot 2023-06-09 at 18 20 43

The expected behavior is to have a background under each line. 

And now comes an additional question which is how border color behaves when you have one line background touching the background from the line below (e.g. padding is larger than the default line spacing) because there can be two states: A or B. What we want, is the B.

Screenshot 2023-06-09 at 18 30 03

src/renderer/canvas/label.ts
```diff
@@ -58,40 +72,160 @@ export class Label {
   }
 }
 
-export const drawLabel = (context: CanvasRenderingContext2D, label: Label) => {
+export const drawLabel = (context: CanvasRenderingContext2D, label: Label, type: 'node' | 'edge') => {
```
tonilastre, line 75: I don't really get why you need information about if it is a `node` or `edge`. Why does it matter?

Both can have multiple lines of text. As far as I can see, you probably wanted to handle the situation where padding is larger than the initial margin and in the node label case, the background color of the text would go over/under the node circle, right?

That's why we have this `textBaseline` property. Nodes have that set as `TOP`, and edges have it as `MIDDLE`. With that info, you have everything you need. No need to send `'node'` and/or `'edge'`.

src/renderer/canvas/label.ts
```diff
@@ -100,11 +234,19 @@ const drawText = (context: CanvasRenderingContext2D, label: Label) => {
...
   context.textBaseline = label.textBaseline;
   context.textAlign = 'center';
-  const lineHeight = label.fontSize * FONT_LINE_SPACING;
 
-  for (let i = 0; i < label.textLines.length; i++) {
-    const line = label.textLines[i];
-    context.fillText(line, label.position.x, label.position.y + i * lineHeight);
+  if (label.textLines.includes('A -> B')) {
+    console.log(borderPoints, label);
+  }
```
tonilastre, line 240: This should be removed. You probably used it for testing.

src/renderer/canvas/label.ts
```diff
@@ -58,40 +72,160 @@ export class Label {
...
+
+  drawTextBackground(context, borderPoints, label);
+  drawTextBackgroundBorder(context, borderPoints, label);
+  drawText(context, borderPoints, label, type);
 };
 
-const drawTextBackground = (context: CanvasRenderingContext2D, label: Label) => {
-  if (!label.properties.fontBackgroundColor || !label.position) {
+const drawTextBackgroundAndBorder = (
```
tonilastre, line 92: The name of this function and what it does doesn't match. It tells that it is doing the drawing, but it is actually getting the points for the background of the text.
When I get into it much deeper, those are actually bounding boxes rather than border points. But this is nitpicking.

src/renderer/canvas/label.ts
```diff
@@ -58,40 +72,160 @@ export class Label {
...
     const line = label.textLines[i];
-    const width = context.measureText(line).width + 2 * margin;
-    context.fillRect(
-      label.position.x - width / 2,
-      label.position.y - baselineHeight - margin + i * lineHeight,
-      width,
-      height,
-    );
+
+    let width = context.measureText(line).width + 2 * margin;
```
tonilastre, line 112: Btw there is a bug in the current ORB, and you might fix it in this PR. The problem is that if you call `measureText` it uses the last set of `context.font` for the measurements.
This means if someone changes the font size, the measurement for the background will be incorrect (it will be calculated on the default font size), but the correct font size will be set only after when the text is being draw with `context.font = label.fontFamily;`

So before doing a `context.measureText` we need to be sure we have a correct font setup, using `context.font = label.fontFamily;`.

src/renderer/canvas/label.ts
```diff
@@ -58,40 +72,160 @@ export class Label {
...
+    borderRadius = width / 2;
+  }
+
+  // ensure that the radius isn't too large for y
+  if (height - 2 * borderRadius < 0) {
+    borderRadius = height / 2;
+  }
+
+  return borderRadius;
+};
```
tonilastre, line 151: This is a cool function that could go to `shapes.ts` utility file, like `getMaxValidBorderRadius`.

src/renderer/canvas/label.ts
```diff
@@ -58,40 +72,160 @@ export class Label {
...
+  context.strokeStyle = borderColor.toString();
+
+  context.beginPath();
+
+  if (borderPoints.length === 1) {
+    const { x, y, height, width } = borderPoints[0];
+    const borderRadius = calculateBorderRadius(fontBorderRadius, width, height);
+    drawRoundRect(context, x, y, width, height, borderRadius);
+  } else {
+    for (let i = 0; i < borderPoints.length; i++) {
```
tonilastre, line 189: I am not 100% sure why you need this logic here. I feel like all this can be done using 3 functions which are: `drawRoundRect` + `context.fill()` and `context.stroke()`.

Here is the idea:

* There are two stages: draw background and draw foreground (the actual text)
* Stage 1: Background
  * baseline height in case if text baseline is `TOP` should incorporate the padding (so it doesn't overlay what is above it, e.g. node circle)
  * for each line of text get the bounding boxes (what you called BorderPoints) - check the `context.font` for `context.measureText` (it is in another comment)
  * for each line call `drawRoundRect` - what is cool is that it will make a path and it will not fill nor stroke it - which is great
  * after all lines are called with `drawRoundRect` we can just do 2 calls: `context.stroke()` (if backgroundBorderWidth > 0)  and `context.fill()` - this fill will override the overlapping strokes.
  
You can test this last piece with this little sample:

```html
<!doctype html>
<html lang="en-US">
  <head>
    <meta charset="utf-8" />
    <title>Canvas</title>
    <style>
      canvas {
        border: 1px solid black;
      }
    </style>
  </head>
  <body>
    <canvas id="tutorial" width="550" height="550"></canvas>
    <script>
      function draw() {
        const canvas = document.getElementById("tutorial");
        if (canvas.getContext) {
          const ctx = canvas.getContext("2d");
          ctx.fillStyle = 'yellow';
          ctx.lineWidth = 1;
          ctx.strokeStyle = 'black';
          ctx.rect(100, 100, 20, 20);
          ctx.rect(110, 110, 20, 20);
          ctx.stroke();
          ctx.fill();
        }
      }
      draw();
    </script>
  </body>
</html>
```

* Stage 2: Foreground
  * Everything should stay the same as before, the only difference is the starting `y` when the text baseline is `TOP` (in the case of the nodes) because of the padding - so the baseline height needs to incorporate the padding.

What do you think?

src/renderer/canvas/label.ts
```diff
@@ -25,6 +32,13 @@ export interface ILabelData {
   properties: Partial<ILabelProperties>;
 }
 
+type BorderPoint = {
```
tonilastre, line 35: Btw, we do all interfaces and types with the prefix `I` so we can differentiate it from the classes with the same name.

tonilastre: But, you already have a type you need, it is called `IRectangle` from `common/rectangle.ts`. Same fields, same types.

src/renderer/canvas/label.ts
shashankshukla96: I think there will be one issue here if there is no backgroundColor i.e. `label.properties.fontBackgroundColor` is not set, in this case with the above method, we will get overlapping borders around the label.

image

So, the border has to be drawn independently of the background.

Please let me know if my understanding is not correct.

src/renderer/canvas/label.ts
shashankshukla96: done

src/renderer/canvas/shapes.ts
```diff
@@ -151,6 +153,8 @@ export const drawRoundRect = (
   w: number,
   h: number,
   r: number,
+  roundedTop: Boolean = true,
+  roundedBottom: Boolean = true,
```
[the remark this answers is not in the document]
shashankshukla96: done

PR #65, opened 2023-06-14
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/orb/actions/runs/5270930886#summary-14267025062

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/orb/actions/runs/5270931012#summary-14267025554

PR #67, opened 2023-07-14
Add check-spelling action

jsoref:
In https://github.com/memgraph/orb/pull/65#pullrequestreview-1530067792, @tonilastre suggested adding the check-spelling workflow.

This is the workflow, roughly as I used it to develop the PR, but with a couple of changes.

As configured check-spelling will produce a job summary reachable via [Details](https://github.com/jsoref/orb/actions/runs/5270930886#summary-14267025062).

When possible, it will generate a Sarif report, and if a user is an admin (or nearly an admin) they'll be able to view it in GitHub, otherwise, users can download a generated artifact and process w/ their own tools (especially VSCode).

I've configured it to provide a fallback message (using `spell_check_this: memgraph/orb@main`) (ideally this shouldn't be necessary, but I think I have a bug I need to resolve which I haven't had time to focus on) -- note that this assumes it'll be merged to main before most people notice the workflow.

The `advice.md` file is a markdown file that the project can customize based on the needs of its maintainers/contributors. The `README.md` is designed to be helpful documentation, but is absolutely optional.

You'll want to review `expect.txt` at least once, and potentially periodically (depending on how good people are at paying attention to the tool -- at work people sometimes just accept its suggestions even when it's pointing out obvious typos 🤷‍♂️). Upon reviewing it while drafting this PR, I noticed a person's name was typo'd (and that I'd effectively missed that in my previous runs).

The update job is designed to only run in forks (see https://github.com/check-spelling/tac/pull/3 for an example of it refusing to run in the configured repository -- sorry, I'm recycling some text, but the same would apply for this workflow...). It enables people to talk to their own instance instead of running the update command locally / manually editing `expect.txt` / `excludes.txt`. You're encouraged to play with this logic in a fork, or if you aren't comfortable with it, there's nothing wrong with omitting the job.

It's using the latest released tag ([v0.0.21](https://github.com/check-spelling/check-spelling/releases/tag/v0.0.21)). Some projects are release-early / release-often -- check-spelling isn't. I don't tend to make a lot of releases although my coworkers are dogfooding [prerelease](https://github.com/check-spelling/check-spelling/tree/prerelease) in a dozen repositories (and I'm constantly dogfooding it to make contributions such as the one that was merged here). If you start using GitHub workflows, you'll want to set up dependabot to make PRs for updates to things.

I'm happy to answer questions here / in Discord, and if this is merged, I'll probably watch the repository for a bit to make sure people don't have any problems.


jsoref:
I'll need to update the workflow. I've made a couple of releases since this PR was opened and I'm sure the code has changed a bit...

PR #69, opened 2023-07-16
New: Add support for enabling and disabling dragging of nodes (fixes #62)

### LINKED ISSUE #62: Option to disable interaction
Is there any way in which I can disable interaction with the graph? No zoom/pan and no drag of nodes 

Thanks

parmar-abhinav:
This PR introduces a new feature that allows users to enable or disable dragging of nodes in the library. By adding a new boolean variable `isDragEnabled` as a configuration option, users can control whether nodes can be dragged or not.

To enable dragging, set `isDragEnabled` to `true`. Nodes will be draggable by default if this option is not provided or set to `false`. When dragging is disabled, users will not be able to move nodes within the graph.

This feature enhances the usability of the library by giving users more control over the interactivity of nodes. It provides flexibility in scenarios where dragging nodes may need to be restricted or enabled based on specific requirements.

### DIFF
--- a/src/views/orb-view.ts
+++ b/src/views/orb-view.ts
@@ -23,11 +23,16 @@ import { SimulatorEventType } from '../simulator/shared';
 import { getDefaultGraphStyle } from '../models/style';
 import { isBoolean } from '../utils/type.utils';
 
+export interface IGraphInteractionSettings {
+  isDragEnabled: boolean;
+}
+
 export interface IOrbViewSettings<N extends INodeBase, E extends IEdgeBase> {
   getPosition?(node: INode<N, E>): IPosition | undefined;
   simulation: Partial<ID3SimulatorEngineSettings>;
   render: Partial<IRendererSettings>;
   strategy: Partial<IEventStrategySettings>;
+  interaction: Partial<IGraphInteractionSettings>;
   zoomFitTransitionMs: number;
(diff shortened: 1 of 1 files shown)


tonilastre (changes requested):
This is cool! Thanks for the contribution.

Checked the code, all good there. Two things are missing though:

* `setSettings` should also listen for the interaction changes. Currently, each settings group is handled separately because each group can have some aftereffects (e.g. rendering, simulation) - you can see it in the `setSettings`. You just need to cover the case when the user wants to override a new group `interaction`.
* Update the docs around the settings that can be updated + default values. You can see the current examples in the `docs/` folder.

parmar-abhinav:
> This is cool! Thanks for the contribution.
> 
> Checked the code, all good there. Two things are missing though:
> 
> * `setSettings` should also listen for the interaction changes. Currently, each settings group is handled separately because each group can have some aftereffects (e.g. rendering, simulation) - you can see it in the `setSettings`. You just need to cover the case when the user wants to override a new group `interaction`.
> * Update the docs around the settings that can be updated + default values. You can see the current examples in the `docs/` folder.

Thank you for your feedback! I have made the necessary updates to address your suggestions. Please check.

tonilastre (approved):

PR #70, opened 2023-07-18
New: Add support for enabling and disabling zoom (fixes #62)

### LINKED ISSUE #62: Option to disable interaction

parmar-abhinav:
**Description:**

This pull request introduces enhancements related to zooming functionality. The changes aim to provide users with more control over zooming interactions. Here's an overview of the modifications made:

- Added a new property `isZoomEnabled` to the optional interaction object.

- `isZoomEnabled` allows users to enable or disable the zooming feature in the graph.

- When `isZoomEnabled` is set to `true`, users can interactively zoom in and out of the graph.

- When `isZoomEnabled` is set to `false`, zooming functionality is disabled, restricting users from changing the zoom level.

**Example:**

```typescript
//To enable zooming
orb.setSettings({ interaction: { isZoomEnabled: true } });
```

```typescript
//To disable zooming
orb.setSettings({ interaction: { isZoomEnabled: false } });
```

### DIFF
--- a/docs/view-default.md
+++ b/docs/view-default.md
@@ -321,15 +321,17 @@ orb.events.on(OrbEventType.MOUSE_CLICK, (event) => {
 
 ### Property `interaction`
 
-The optional property `interaction` has one property that you can enable/disable:
+The optional property `interaction` has two properties that you can enable/disable:
 
 * `isDragEnabled` - property controls the dragging behavior within the application. When it is set to `true`, dragging is enabled, allowing users to interact with nodes and edges by dragging them to different positions within the graph. On the other hand, when `isDragEnabled`` is set to false, dragging functionality is disabled, preventing users from moving or repositioning nodes and edges through dragging interactions.
 
-This property provides a straightforward way to enable or disable the dragging feature based on the needs and requirements of your application. By toggling the value of `isDragEnabled`, you can easily control whether users are allowed to interactively reposition elements within the graph by dragging them. e.g:
+* `isZoomEnabled` - This property controls the zooming behavior within the application. Setting it to `true` enables zooming, allowing users to interactively zoom in and out of the graph. Setting it to `false` disables zooming, restricting the user's ability to change the zoom level.
+
+These properties provide a straightforward way to enable or disable dragging and zooming features based on the needs and requirements of your application. By toggling the values of isDragEnabled and isZoomEnabled, you can easily control the interactivity options available to users. e.g:
 
 ```typescript
-// Disable default drag interaction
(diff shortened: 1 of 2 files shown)


docs/view-default.md
```diff
@@ -321,15 +321,17 @@ orb.events.on(OrbEventType.MOUSE_CLICK, (event) => {
 
 ### Property `interaction`
 
-The optional property `interaction` has one property that you can enable/disable:
+The optional property `interaction` has two properties that you can enable/disable:
```
tonilastre, line 324: In this file, lines `:157` and `:85` are missing the mention of the `isZoomEnabled`.

src/views/orb-view.ts
```diff
@@ -346,6 +355,10 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
   };
 
   zoomed = (event: D3ZoomEvent<any, any>) => {
+    // If zoom is disabled then return
```
tonilastre, line 358: Btw it wasn't required for the drag because nodes can't be dragged, but the zoom interaction disable/enable should be implemented on the `orb-view-map`. This can be another PR.

tonilastre (changes requested):
A small note in the docs missing. Everything else is good!

docs/view-default.md
parmar-abhinav: Added `isZoomEnabled` pls check

src/views/orb-view.ts
parmar-abhinav: Sure will raise a new PR related to changes in `orb-view-map`.

src/views/orb-view.ts
```diff
@@ -230,6 +232,13 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
         this._settings.interaction.isDragEnabled =
           settings.interaction.isDragEnabled;
       }
+
+      // Check if isZoomEnabled is a boolean value
```
tonilastre, line 236: Btw you can remove these comments (including the ones for the drag check). It is usually good practice to just have a comment when you need to describe a complex part of the code. The code here is already well written so it doesn't need comments.

tonilastre (changes requested):
There is one additional bug with the recenter + zoom. Check the suggestion and other comments regarding inline comments.

Btw, I am not sure why `husky` wasn't working on your machine, but please run `npm run lint:fix` because there are some lint issues in the code :) This will automatically fix them so you can commit them afterward.

Issue #68, opened 2023-07-16
Any way to select only node/edge which is clicked


parmar-abhinav:
When any node is selected then its adjacent edges/nodes also gets selected. passing `isDefaultSelectEnabled` to false in strategy does not select any node/edge.

any way to select only the node/edge which is clicked ?

tonilastre:
I totally agree about providing greater customization based on specific use cases. I think the `selectClickedOnly` would just add a bit to the default strategy, but it wouldn't solve the general problem: How to enable users of the Orb API to be able to do any kind of select/hover flows. E.g.:

* Select only the clicked node
* Select the clicked node, but also select the adjacent nodes
* Select the path between two clicked nodes
* Upon clicking, select the clicked node on the first click, and then select 1 hop more for each new click on the same node - more like spread clicking

There are so many cases and we can't cover them all with flags, so my thinking is in this direction:

* The ORB should have a structure to keep nodes/edges by their state (selected, hovered, or even custom user ones - that is the reason why the state can be any number) - this is for easier and faster lookup, currently, you need to iterate through all nodes/edges to get the selected/hovered one which is not an optimal way
* The state change of the node/edge goes through `setState` which will be listened to by ORB and ORB components (renderer). ORB will react to it by updating the internal structure for faster lookup, while the renderer should do the throttling for the `render` and `simulate` calls - this requires integration of `RxJS` which we have in plan and which will happen on the `release/1.0.0` branch

With just these above two changes, the `default strategy` will be questionable how much sense it has. There will be two paths:

1) If you want custom select/hover logic, you disable the default via settings and implement yours in the event listeners. (like the example above)

2) If you want custom select/hover logic, you will implement a custom callback for each select/hover settings function - I am talking about the functions from the `DefaultEventStrategy` class: https://github.com/memgraph/orb/blob/main/src/models/strategy.ts - these will be exposed out in the settings.

Additionally, regardless of the two paths (event listener or callback implementation), an update on the `setState` for nodes and edges can be done with optional options, e.g.

```typescript
// Select the node
node.setState({ state: SELECTED });

// Select the node if not selected, otherwise unselect it
node.setState({ state: SELECTED, options: { isToggle: true }});

// Select the node, but unselect any other node that is currently selected (aka only the single node will be selected)
node.setState({ state: SELECTED, options: { isSingle: true }}};
```

What do you think?

parmar-abhinav:
[...]

PR #71, opened 2023-07-23
New: Added support to set node/edge state using custom options

parmar-abhinav:
Issue id: #68 

**Description:**
This pull request introduces the mechanism for handling state changes of nodes and edges through the `setState` method in the Node and Edge classes. Additionally, this enhancement includes optional parameters such as `isToggle` and `isSingle` in the `setState` method, providing greater customization for handling state updates.

**Examples:**
```typescript
// Select the node
node.setState({ state: SELECTED });

// Select the node if not selected, otherwise unselect it
node.setState({ state: SELECTED, options: { isToggle: true }});

// Select the node, but unselect any other node that is currently selected (aka only the single node will be selected)
node.setState({ state: SELECTED, options: { isSingle: true }}};
```

### DIFF
--- a/src/models/edge.ts
+++ b/src/models/edge.ts
@@ -1,5 +1,5 @@
 import { INodeBase, INode } from './node';
-import { GraphObjectState } from './state';
+import { GraphObjectState, ISetStateOptions } from './state';
 import { Color, IPosition, ICircle, getDistanceToLine } from '../common';
 
 const CURVED_CONTROL_POINT_OFFSET_MIN_SIZE = 4;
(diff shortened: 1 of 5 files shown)


parmar-abhinav:
This PR contain only one part of the 2 task mentioned in the issue. Implementing `setState` method in this PR. Thinking of raising separate PR for the task:

> The ORB should have a structure to keep nodes/edges by their state (selected, hovered, or even custom user ones - that is the reason why the state can be any number) - this is for easier and faster lookup, currently, you need to iterate through all nodes/edges to get the selected/hovered one which is not an optimal way

PR #74, opened 2023-08-01
New: Add zoom and recenter functions

cizl:


src/views/orb-view.ts
```diff
@@ -300,6 +300,18 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
 
+  zoomIn(onRendered?: () => void) {
+    select(this._canvas)
+      .transition()
+      .duration(this._settings.zoomFitTransitionMs)
+      .ease(easeLinear)
+      .call(this._d3Zoom.scaleBy, 1.2)
+      .call(() => {
+        this._renderer.render(this._graph);
+        onRendered?.();
```
tonilastre, line 311: You could call `this.render(onRendered)` here too, sam as for the `zoomOut`.

tonilastre (approved):

PR #76, opened 2023-08-09
New: Added support for keyboard interaction

parmar-abhinav:
This PR introduces and new feature that allows user to Drag/Zoom graph in Canvas using `keyboard`. Allowing users to control the scale of zoom using `zoomInFactor` and `zoomOutFactor` also `draggingFactor` allows to control dragging speed. This feature enhances the usability of the library by giving users more control over interactivity.

### DIFF
--- a/docs/view-default.md
+++ b/docs/view-default.md
@@ -84,6 +84,12 @@ interface IOrbViewSettings {
   interaction: {
     isDragEnabled: boolean;
     isZoomEnabled: boolean;
+    keyboard: {
+      isKeyboardEnabled: boolean;
+      zoomInFactor: number;
+      zoomOutFactor: number;
+      draggingFactor: number;
+    };
   };
   // Other default view parameters
   zoomFitTransitionMs: number;
@@ -157,6 +163,12 @@ const defaultSettings = {
   interaction: {
     isDragEnabled: true;
     isZoomEnabled:  true;
+    keyboard: {
+      isKeyboardEnabled: true,
+      zoomInFactor: 1.2,
+      zoomOutFactor: 0.8,
+      draggingFactor: 25
+    }
   },
   zoomFitTransitionMs: 200,
   isOutOfBoundsDragEnabled: false,
@@ -323,13 +335,19 @@ orb.events.on(OrbEventType.MOUSE_CLICK, (event) => {
 
 ### Property `interaction`
 
-The optional property `interaction` has two properties that you can enable/disable:
+The `interaction` property controls the interactivity options available to users when interacting with the graph visualization.
 
 * `isDragEnabled` - property controls the dragging behavior within the application. When it is set to `true`, dragging is enabled, allowing users to interact with nodes and edges by dragging them to different positions within the graph. On the other hand, when `isDragEnabled`` is set to false, dragging functionality is disabled, preventing users from moving or repositioning nodes and edges through dragging interactions.
 
 * `isZoomEnabled` - This property controls the zooming behavior within the application. Setting it to `true` enables zooming, allowing users to interactively zoom in and out of the graph. Setting it to `false` disables zooming, restricting the user's ability to change the zoom level.
 
-These properties provide a straightforward way to enable or disable dragging and zooming features based on the needs and requirements of your application. By toggling the values of isDragEnabled and isZoomEnabled, you can easily control the interactivity options available to users. e.g:
+* `keyboard` - This property allows you to define keyboard interaction settings:
+   * `isKeyboardEnabled` (boolean): Set this property to `true` to enable keyboard interaction within the graph visualization. When enabled, users can use keyboard shortcuts for certain interactions.
+   * `zoomInFactor` (number): The zoom-in factor determines the extent to which the graph will zoom in when the user presses the zoom-in keyboard shortcut.
+   * `zoomOutFactor` (number): The zoom-out factor determines the extent to which the graph will zoom out when the user presses the zoom-out keyboard shortcut.
+   * `draggingFactor` (number): The dragging factor specifies the speed at which nodes and edges will move when the user interacts with them using keyboard shortcuts for dragging.
+
+By customizing the interaction property, you can enable or disable dragging and zooming features based on the needs and requirements of your graph visualization application. Additionally, you can also choose to enable keyboard interactions for further user control and convenience. e.g:
 
 ```typescript
 // Disable default drag interaction and enable zooming
--- a/src/renderer/canvas/canvas-renderer.ts
+++ b/src/renderer/canvas/canvas-renderer.ts
@@ -245,6 +245,47 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
     return zoomIdentity.translate(newX, newY).scale(newZoom);
   }
 
+  getZoomTransform(zoomFactor: number): ZoomTransform {
+    const previousZoom = this.transform.k;
+    const newZoom = Math.max(
+      Math.min(zoomFactor * previousZoom, this._settings.maxZoom),
+      this._settings.minZoom
+    );
+    return zoomIdentity.scale(newZoom);
+  }
+
+  getDragLeftTransform(draggingFactor: number): ZoomTransform {
+    const currentTransform = this.transform;
+    const newX = currentTransform.x + draggingFactor;
+    return zoomIdentity
+      .translate(newX, currentTransform.y)
+      .scale(currentTransform.k);
+  }
+
+  getDragRightTransform(draggingFactor: number): ZoomTransform {
+    const currentTransform = this.transform;
+    const newX = currentTransform.x - draggingFactor;
+    return zoomIdentity
+      .translate(newX, currentTransform.y)
+      .scale(currentTransform.k);
+  }
+
+  getDragUpTransform(draggingFactor: number): ZoomTransform {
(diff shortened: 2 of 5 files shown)


src/renderer/canvas/canvas-renderer.ts
```diff
@@ -245,6 +245,47 @@ export class CanvasRenderer<N extends INodeBase, E extends IEdgeBase> extends Em
...
+      .scale(currentTransform.k);
+  }
+
+  getDragDownTransform(draggingFactor: number): ZoomTransform {
+    const currentTransform = this.transform;
+    const newY = currentTransform.y - draggingFactor;
+    return zoomIdentity
+      .translate(currentTransform.x, newY)
+      .scale(currentTransform.k);
+  }
```
tonilastre, line 287: All these are pretty much doing the same thing, so I would propose having one function that handles all 4 cases with enum, e.g. `PanDirectionType.{UP, DOWN, LEFT, RIGHT}`.

Btw this event should be `pan`, `drag` is used for dragging nodes/edges. Pan is when you move the canvas up/down/left/right.
With the enum, you can have just one function `getPanTransform(panDirectionType, factor);`

src/views/orb-view.ts
```diff
@@ -364,6 +390,80 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
     }, 1);
   };
 
+  _handleKeyDown = (event: KeyboardEvent) => {
```
tonilastre, line 393: This is missing on the `orb-map-view`.

tonilastre (changes requested):
Nice feature!

docs/view-default.md
```diff
@@ -157,6 +163,12 @@ const defaultSettings = {
   interaction: {
     isDragEnabled: true;
     isZoomEnabled:  true;
+    keyboard: {
+      isKeyboardEnabled: true,
```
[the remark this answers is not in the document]
tonilastre: Also, you already have a group name `keyboard`, so this field can just be `isEnabled`.

src/views/orb-view.ts
```diff
@@ -364,6 +390,80 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+  _dragByArrowKey = (transform: ZoomTransform) => {
+    select(this._canvas)
+      .transition()
+      .duration(this._settings.zoomFitTransitionMs)
+      .ease(easeLinear)
+      .call(this._d3Zoom.transform, transform)
+      .call(() => {
+        this._renderer.render(this._graph);
+      });
+  };
```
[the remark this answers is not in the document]
parmar-abhinav: Thank you for your feedback. I've made the necessary changes based on your suggestions. Pls check.

docs/view-default.md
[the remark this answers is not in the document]
parmar-abhinav: Yes by default it is `false`. Renamed `isKeyboardEnabled` to `isEnabled` and also update default value

src/renderer/canvas/canvas-renderer.ts
parmar-abhinav: Created single `getPanTransform` function, Pls check.

src/views/orb-view.ts
parmar-abhinav: "orb-map-view" is already responding to key presses, so I decided not to make any changes.

src/views/orb-view.ts
```diff
@@ -364,6 +384,73 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+
+  _zoomIn = (zoomInFactor: number) => {
+    const transform = this._renderer.getZoomTransform(zoomInFactor);
+    this._d3Zoom.scaleTo(select(this._canvas), transform.k);
+  };
+
+  _applyTransformation = (transform: ZoomTransform, callback?: () => void) => {
+    select(this._canvas)
+      .transition()
+      .duration(this._settings.zoomFitTransitionMs)
```
tonilastre, line 445: Sorry, I totally missed this one in the last PR check. This `this._settings.zoomFitTransitionMs` is actually from the `recenter` function that does fit to zoom.

In order to make this `applyTransformation` generic enough, I propose having a second argument `options` (that will currently have `transitionMs` (+ callback) only. Something like this:

```typescript
interface IApplyTransitionOptions {
  transitionMs: number;
  callback: () => void;
}

_applyTransformation = (transform: ZoomTransform, options?: Partial<IApplyTransitionOptions>) => {
  const transitionMs = options?.transitionMs ?? DEFAULT_TRANSITION_MS;
  ...
```

Then, the recenter can send its `this._settings.zoomFitTransitionMs` and `callback`. The zoom, and pan can send their `transitionMs`. This `transitionMs` can be even added to the settings with a default value that is lower than 200, maybe 100 or 50. 200 seems a bit slow for expected keyboard movement.

src/views/orb-view.ts
```diff
@@ -150,6 +168,8 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
       .on('contextmenu', this.mouseRightClicked)
       .on('dblclick.zoom', this.mouseDoubleClicked);
 
+    document.addEventListener('keydown', this._handleKeyDown);
```
tonilastre, line 171: Ok, this is a bit problematic, to listen for the key down events on the whole document. If there is a form next to the Orb view, orb will get each event and even react to +, -, array keys while the user writes keys in the input.

So, this should be fixed either by:

* Listening only for keys on the container - I don't think this is even possible
* Having a way to listen for click events in the container to make it "focused" so we can know if we want to react to document.keydown event (for example, check the `orb-view-map` and how it works - key events are working as expected there: if you click on a map, it becomes "focused" in the background, and keys are working. If you click on the header, keys are not working for the graph view out of the orb view, which is the expected behavior). 

If you have any other suggestions on how to approach this problem, feel free to share. Happy to discuss this.

src/views/orb-view.ts
tonilastre: Awesome! Yep, that makes sense.

tonilastre (changes requested):
It is better and more performant code now! Btw I have some comments mostly about moving stuff to other files, missing property, and handling key events globally.

src/views/orb-view.ts
```diff
@@ -364,6 +384,73 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+      }
+      case 'ArrowDown': {
+        const panFactor = this._settings.interaction.keyboard.panFactor;
+        const dragDownTransform = this._renderer.getPanTransform(PanDirectionType.DOWN, panFactor);
+        this._applyTransformation(dragDownTransform);
+        break;
+      }
+      default:
+        break;
+    }
```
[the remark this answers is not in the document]
parmar-abhinav: I've made the recommended changes to improve the code readability and reduce redundancy.

src/views/orb-view.ts
parmar-abhinav: I've implemented your suggested approach to enhance the generic nature of the `applyTransformation` function. 

Thank you for your valuable input.

src/views/orb-view.ts
```diff
@@ -26,8 +32,20 @@ import { isBoolean } from '../utils/type.utils';
...
+    zoomOutFactor: number;
+    panFactor: number;
+    transitionMs: number;
+  };
 }
 
+const DEFAULT_ZOOM_IN_FACTOR = 1.2;
+const DEFAULT_ZOOM_OUT_FACTOR = 0.8;
+const DEFAULT_PAN_FACTOR = 25;
+const DEFAULT_TRANSITION_MS = 200;
```
tonilastre, line 47: Maybe the default can be `0` - no transition.

src/views/orb-view.ts
tonilastre: @parmar-abhinav do you have any info on this?

tonilastre (changes requested):
@parmar-abhinav everything looks great, there are just two comments left: one for default and another one for document-level listening (which might be problematic when integrating orb). Please check those two and let me know.

src/views/orb-view.ts
parmar-abhinav: Done, Please check

src/views/orb-view.ts
parmar-abhinav: Tried few approach i.e.
1. Instead of attaching the `keydown` event listener to the entire document. I tried attaching it to the specific container where we want to handle the key events. But not able to listen to `keydown` event using this approach.
```typescript
this._container.addEventListener('keydown', (event) => {})
```

2. Also tried listening to `focusin` and `focusout` event on the container. But it is also not working.
```typescript
private readonly isContainerFocused = false;

this._container.addEventListener('focusin', () => {
    this.isContainerFocused = true;
});

this._container.addEventListener('focusout', () => {
    this.isContainerFocused = false;
});

document.addEventListener('keydown', event => {
    if (this.isContainerFocused) {
        // Handle keydown events only when the container is focused
        this._handleKeyDown(event);
    }
});

```

tonilastre: Hmm, thanks for sharing. I will check how the leaflet did it (it is used for orb-view-map), maybe we can replicate that behavior. I will share my findings here.

PR #77, opened 2023-08-09
New Added support for custom edge line style

parmar-abhinav:
This PR introduces a new feature allowing users to have greater control over the appearance of edges in the graph by implementing custom edge line styles. User can set a `Solid`, `Dashed`, `Dotted` or even `custom` edges by defining own edge patterns.

### DIFF
--- a/src/models/edge.ts
+++ b/src/models/edge.ts
@@ -1,6 +1,7 @@
 import { INodeBase, INode } from './node';
 import { GraphObjectState } from './state';
 import { Color, IPosition, ICircle, getDistanceToLine } from '../common';
+import { isArrayOfNumbers } from '../utils/type.utils';
 
 const CURVED_CONTROL_POINT_OFFSET_MIN_SIZE = 4;
 const CURVED_CONTROL_POINT_OFFSET_MULTIPLIER = 4;
@@ -25,6 +26,18 @@ export interface IEdgePosition {
   target: any;
 }
 
+export enum EdgeLineStyleType {
+  SOLID = "solid",
+  DASHED = "dashed",
+  DOTTED = "dotted",
+  CUSTOM = "custom",
+}
+
+export interface IEdgeLineStyle {
+  type: EdgeLineStyleType;
+  dashPattern: number[];
+}
+
 /**
  * Edge style properties used to style the edge (color, width, label, etc.).
  */
@@ -46,6 +59,7 @@ export type IEdgeStyle = Partial<{
   widthHover: number;
   widthSelected: number;
   zIndex: number;
+  edgeLineStyle: IEdgeLineStyle;
 }>;
 
 export interface IEdgeData<N extends INodeBase, E extends IEdgeBase> {
@@ -89,6 +103,7 @@ export interface IEdge<N extends INodeBase, E extends IEdgeBase> {
   hasShadow(): boolean;
   getWidth(): number;
   getColor(): Color | string | undefined;
+  getEdgeLineStyle(): IEdgeLineStyle;
 }
 
 export class EdgeFactory {
@@ -255,6 +270,20 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> implements IEdge<N
     }
 
     return color;
+  }
+
+  getEdgeLineStyle(): IEdgeLineStyle {
+    const edgeLineStyle: IEdgeLineStyle = {
+      type: EdgeLineStyleType.SOLID,
+      dashPattern: [],
+    };
+    if (this.style.edgeLineStyle !== undefined) {
+      edgeLineStyle.type = this.style.edgeLineStyle.type;
+      if (isArrayOfNumbers(this.style.edgeLineStyle.dashPattern)) {
+        edgeLineStyle.dashPattern = this.style.edgeLineStyle.dashPattern;
+      }
+    }
+    return edgeLineStyle;
   }
 }
 
--- a/src/renderer/canvas/edge/shared.ts
+++ b/src/renderer/canvas/edge/shared.ts
@@ -10,3 +10,7 @@ export interface IEdgeArrow {
   angle: number;
   length: number;
 }
+
+export const DEFAULT_SOLID_LINE_PATTERN: number[] = [];
+export const DEFAULT_DASHED_LINE_PATTERN: number[] = [5, 5];
+export const DEFAULT_DOTTED_LINE_PATTERN: number[] = [1, 1];
--- a/src/renderer/canvas/edge/types/edge-straight.ts
+++ b/src/renderer/canvas/edge/types/edge-straight.ts
@@ -1,6 +1,16 @@
 import { INodeBase, INode } from '../../../../models/node';
-import { EdgeStraight, IEdgeBase } from '../../../../models/edge';
-import { IBorderPosition, IEdgeArrow } from '../shared';
+import {
+  EdgeLineStyleType,
+  EdgeStraight,
+  IEdgeBase,
+} from "../../../../models/edge";
+import {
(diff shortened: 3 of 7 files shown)


src/renderer/canvas/edge/types/edge-straight.ts
```diff
@@ -17,6 +27,28 @@ export const drawStraightLine = <N extends INodeBase, E extends IEdgeBase>(
...
+      break;
+    case EdgeLineStyleType.CUSTOM: {
+      const dashPattern: number[] = edge.getEdgeLineStyle().dashPattern;
+      context.setLineDash(dashPattern);
+      break;
+    }
+    default:
+      context.setLineDash(DEFAULT_SOLID_LINE_PATTERN);
+      break;
+  }
```
tonilastre, line 50: This is repeated in all three types, but it is all the same code. Additionally, rendering should be super fast which means all unnecessary calls should be removed. For example, for each edge calling `getEdgeLineStyle` creates a new object just to figure out the correct style - when rendering has a min of 25 FPS, this will start to impact the process.

I would suggest removing `getEdgeLineStyle` - no need for it because you have the latest information in the `this.style` of the edge model.
I would suggest having new function on the edge model, something like `getLineDashPattern` that returns `number[] | null`. In case of solid or undefined lineStyle, it returns null. In case of any other, it returns a default one (for dashed or dotted) or a user-defined one (in case of custom). 

Then this code (and in all three edge types) would be simple as:

```
const lineDashPattern = edge.getLineDashPattern();
context.setLineDashPatern(lineDashPattern ?? []);
```

src/renderer/canvas/edge/shared.ts
```diff
@@ -10,3 +10,7 @@ export interface IEdgeArrow {
   angle: number;
   length: number;
 }
+
+export const DEFAULT_SOLID_LINE_PATTERN: number[] = [];
+export const DEFAULT_DASHED_LINE_PATTERN: number[] = [5, 5];
+export const DEFAULT_DOTTED_LINE_PATTERN: number[] = [1, 1];
```
tonilastre, line 16: Maybe you will need to move this closer to the `models/edge` - check the other comment on where the code around patterns should be.

src/utils/type.utils.ts
```diff
@@ -87,3 +87,18 @@ export const isNull = (value: any): value is null => {
...
   return typeof value === 'function';
 };
+
+/**
+ * Type check for array of numbers
+ *
+ * @param {any} value Any value
+ * @return {boolean} True if it is a array of numbers, false otherwise
+ */
+export const isArrayOfNumbers = (value: any): value is boolean => {
```
tonilastre, line 97:
```suggestion
export const isArrayOfNumbers = (value: any): value is number[] => {
```

In typeguards, this param should return the correct type. Typeguards return a boolean by default + this is for the Typescript to be able to know what the type is after typeguard returns true.

src/utils/type.utils.ts
```diff
@@ -87,3 +87,18 @@ export const isNull = (value: any): value is null => {
...
+  if (!isArray(value)) {
+    return false;
+  }
+
+  const isNotNumber = value.some((element) => !isNumber(element));
+  return !isNotNumber;
```
tonilastre, line 103:
```suggestion
  return value.some((element) => !isNumber(element));
```

This is pretty much clear enough where you don't have any benefit from an additional variable. If this was a more complex function, then the named variable would be a good choice for better readability.

src/models/edge.ts
```diff
@@ -46,6 +59,7 @@ export type IEdgeStyle = Partial<{
   widthHover: number;
   widthSelected: number;
   zIndex: number;
+  edgeLineStyle: IEdgeLineStyle;
```
tonilastre, line 62: No need to have `edge` because you already have a context of `EdgeStyle`. I would suggest naming this `lineStyle`.

src/models/edge.ts
```diff
@@ -25,6 +26,18 @@ export interface IEdgePosition {
...
+  SOLID = "solid",
+  DASHED = "dashed",
+  DOTTED = "dotted",
+  CUSTOM = "custom",
+}
+
+export interface IEdgeLineStyle {
+  type: EdgeLineStyleType;
+  dashPattern: number[];
+}
```
tonilastre, line 39: This should be fixed from a typescript side to be more precise. For example, currently, you can have `{ type: "solid", dashPattern: [1, 2, 3] }` which is not correct. An array can only be used with `custom` type. E.g.

```typescript
export type IEdgeLineStyle =
  | { type: EdgeLineStyleType.SOLID }
  | { type: EdgeLineStyleType.DASHED }
  | { type: EdgeLineStyleType.DOTTED }
  | { type: EdgeLineStyleType.CUSTOM, pattern: number[] };
```

or even

```typescript
export type IEdgeLineStyle =
  | EdgeLineStyleType.SOLID
  | EdgeLineStyleType.DASHED
  | EdgeLineStyleType.DOTTED
  | { type: EdgeLineStyleType.CUSTOM, pattern: number[] };
```

Maybe the second one might even work better if users will use SOLID, DASHED or DOTTED more. The first one is easier to use from library point of view, the second one from user point of view.

tonilastre: Nitpicking but I am also thinking about renaming `dashPattern` to just `pattern` because you already have a context which is a line style. What do you think? Also, what do you think about those two proposals above, which one makes more sense to you?

tonilastre (changes requested):

src/renderer/canvas/edge/types/edge-straight.ts
parmar-abhinav: Thanks for your feedback. I've made the following changes to the code as you suggested:

- I removed the getEdgeLineStyle() function, as you pointed out that it is unnecessary and creates a new object each time it is called.

- I added a new function called getLineDashPattern() to the edge model. This function returns a number[] or null, depending on whether the edge has a custom line style or not.

- I updated the code to use the getLineDashPattern() function instead of getEdgeLineStyle().

I believe these changes will make the code more efficient and improve the rendering performance.

src/models/edge.ts
parmar-abhinav: Renamed it to `lineStyle`, pls check.

src/models/edge.ts
parmar-abhinav: I agree with your suggestion to fix the TypeScript type for `IEdgeLineStyle`.

Yes The second approach might even work better if users will use SOLID, DASHED or DOTTED more. However, I am thinking to use the first approach because it gives users a more consistent and predictable experience. With the first approach, users know that they can always specify the line style using the same syntax, regardless of whether they are using a predefined style or a custom style.

tonilastre: Yep, agree with you. The first approach is slightly a better option because of the usage predictability.

src/models/edge.ts
```diff
@@ -256,6 +274,25 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> implements IEdge<N
...
+      return null;
+    }
+
+    switch (lineStyle.type) {
+      case EdgeLineStyleType.DASHED:
+        return DEFAULT_DASHED_LINE_PATTERN;
+      case EdgeLineStyleType.DOTTED:
+        return DEFAULT_DOTTED_LINE_PATTERN;
+      case EdgeLineStyleType.CUSTOM:
+        return isArrayOfNumbers(lineStyle.pattern) ? lineStyle.pattern : null;
```
tonilastre, line 291: Nothing to change here, but just a general comment that once node and edge style have a setter for styles `setStyle` (what 1.0.0 will require), this check will be done there because the check is done once then.

tonilastre (approved):

PR #78, opened 2023-08-09
New Added support for node label alignment

parmar-abhinav:
This PR introduces a new feature allowing users to have greater control over the node label alignment in the graph by defining custom not label positions i.e. top, left, right, bottom

### DIFF
--- a/src/renderer/canvas/node.ts
+++ b/src/renderer/canvas/node.ts
@@ -1,7 +1,20 @@
-import { INodeBase, INode, NodeShapeType } from '../../models/node';
-import { IEdgeBase } from '../../models/edge';
-import { drawDiamond, drawHexagon, drawSquare, drawStar, drawTriangleDown, drawTriangleUp, drawCircle } from './shapes';
-import { drawLabel, Label, LabelTextBaseline } from './label';
+import {
+  INodeBase,
+  INode,
+  NodeShapeType,
+  NodeLabelAligment,
+} from "../../models/node";
+import { IEdgeBase } from "../../models/edge";
+import {
+  drawDiamond,
+  drawHexagon,
+  drawSquare,
+  drawStar,
+  drawTriangleDown,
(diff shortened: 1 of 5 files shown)


parmar-abhinav:
@tonilastre Can you pls review the PR.

src/renderer/canvas/node.ts
```diff
@@ -97,9 +110,43 @@ const drawNodeLabel = <N extends INodeBase, E extends IEdgeBase>(
...
+      labelX += distance;
+      labelTextAlign = LabelTextAlign.LEFT;
+      labelTextBaseline = LabelTextBaseline.MIDDLE;
+      break;
+    default:
+      labelY += distance;
+      labelTextAlign = LabelTextAlign.CENTER;
+      labelTextBaseline = LabelTextBaseline.TOP;
+      break;
+  }
```
tonilastre, line 144: Actually, this should be part of Label logic. I would propose the following:

```
NodeLabelPosition.{TOP, BOTTOM, LEFT, RIGHT} // default = BOTTOM
NodeLabelAlignment.{LEFT, RIGHT, CENTER} // default = CENTER

EdgeLabelPosition.{CENTER} // default = CENTER
EdgeLabelAlignment.{LEFT, RIGHT, CENTER} // default = CENTER
```

And then Label logic receives position + alignment and handles the rest. Please take care of multiple lines because it is not working well. Check the image below for the "top".

Screenshot 2023-08-16 at 22 22 31

tonilastre: Node logic should just handle the start position: `x`, `y` and give that to the Label logic. Label should take the start position + label position + label alignment + text lines to figure out where each line should be.

tonilastre (changes requested):
This is a really cool feature, but there are some issues (e.g. `top`, or combining `left` vs `right`). Please check the comments.

PR #80, opened 2023-08-25
feat: added optional instance setting to allow consumers to choose si…

eacaps:
I'm not thrilled with how this turned out, and I think the typing of `IOrbViewSettingsInit` might be slightly off now, so please let me know if you'd like me to make adjustments

### DIFF
--- a/src/views/orb-view.ts
+++ b/src/views/orb-view.ts
@@ -44,7 +44,11 @@ export interface IOrbViewSettings<N extends INodeBase, E extends IEdgeBase> {
 export type IOrbViewSettingsInit<N extends INodeBase, E extends IEdgeBase> = Omit<
   Partial<IOrbViewSettings<N, E>>,
   'render'
-> & { render?: Partial<IRendererSettingsInit> };
+> & { render?: Partial<IRendererSettingsInit> } & { 
+  instances: {
+    simulator?: ISimulator | (() => ISimulator);
+  };
+};
 
 export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbView<N, E, IOrbViewSettings<N, E>> {
   private _container: HTMLElement;
@@ -150,7 +154,7 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
       .on('contextmenu', this.mouseRightClicked)
       .on('dblclick.zoom', this.mouseDoubleClicked);
 
-    this._simulator = SimulatorFactory.getSimulator();
+    this._simulator = this._createSimulator(settings?.instances?.simulator);
     this._simulator.on(SimulatorEventType.SIMULATION_START, () => {
       this._isSimulating = true;
       this._simulationStartedAt = Date.now();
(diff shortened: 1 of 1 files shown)


src/views/orb-view.ts
```diff
@@ -44,7 +44,11 @@ export interface IOrbViewSettings<N extends INodeBase, E extends IEdgeBase> {
 export type IOrbViewSettingsInit<N extends INodeBase, E extends IEdgeBase> = Omit<
   Partial<IOrbViewSettings<N, E>>,
   'render'
-> & { render?: Partial<IRendererSettingsInit> };
+> & { render?: Partial<IRendererSettingsInit> } & { 
+  instances: {
+    simulator?: ISimulator | (() => ISimulator);
+  };
```
tonilastre, line 50:
```suggestion
> & {
  render?: Partial<IRendererSettingsInit>,
  instances?: {
    simulator?: ISimulator | (() => ISimulator);
  };
```

I would say that you don't need to join them with `&` - just append to the `render` interface. Btw `instances` should be optional, if not users would need to always initiate an empty object even if they don't want to override the simulator instance.

src/views/orb-view.ts
```diff
@@ -569,6 +573,12 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
     this._simulator.simulate();
   }
 
+
+private _createSimulator(simulator: ISimulator | (() => ISimulator) | undefined): ISimulator {
+  if(typeof simulator === "function")  return simulator();
+  return simulator || SimulatorFactory.getSimulator();
+}
```
tonilastre, line 580: Hmm, I think lint is not working for you because this should be indented + if should be wrapped in brackets. Can you check if husky is initialized and `npm run lint` / `npm run lint:fix` works.

Also, there is a cool Typescript trick you can use here:

```typescript
type ICallbackOrInstance<T> = T | (() => T);

export type IOrbViewSettingsInit<N extends INodeBase, E extends IEdgeBase> = Omit<
  ...
  instances?: {
    simulator?: ICallbackOrInstance<ISimulator>;
  };
}


private _createSimulator(simulator?: ICallbackOrInstance<ISimulator>): ISimulator {
   ...
}
```

tonilastre (changes requested):
Looks great! I've just added a few comments regarding the Typescript and lint. Let me know what you think.

src/views/orb-view.ts
eacaps: yeah that definitely looks nicer. is there a cleaner way to do the type check instead of
```
 if(typeof simulator === "function")  return simulator();
 ```

tonilastre: I think that is the most clean way to check for the function. You could have a util function somewhere. Just as I wrote this comment to give you a link to which place is the best to have a utility function in, I found a function:

https://github.com/memgraph/orb/blob/main/src/utils/type.utils.ts#L87

You already have `isFunction` type guard check so you can use it here :)

PR #82, opened 2023-12-14
Fix: resizeObserver is not unobserved (fixes #81)

### LINKED ISSUE #81: MapView/DefaultView: resizeObserver not removed
I have seen that the ResizeObserver inside the views is not cleaned up properly.

In src/views/map-view.ts in the MapView class the following resizeObserver is created:

```
// Resize the canvas based on the dimensions of it's parent container <div>.
const resizeObs = new ResizeObserver(() => this._handleResize());
resizeObs.observe(this._container);
this._handleResize();
```

Similar in src/views/default-view.ts

In my opinion, a call to `resizeObs.unobserve()` is missing, when the `destroy()` method is called, and the observer has to put as a private class variable.

Currently, I receive following error after the graph instance has been destroyed:
```
leaflet-src.js:2456 Uncaught TypeError: Cannot read properties of undefined (reading '_leaflet_pos')
    at getPosition (leaflet-src.js:2456:14)
    at NewClass._getMapPanePos (leaflet-src.js:4455:12)
    at NewClass._rawPanBy (leaflet-src.js:4233:37)
    at NewClass.invalidateSize (leaflet-src.js:3548:12)
    at MapView._handleResize (map-view.js:304:27)
    at ResizeObserver.<anonymous> (map-view.js:42:57)
```

markusnissl:
Fixes #81

### DIFF
--- a/src/views/default-view.ts
+++ b/src/views/default-view.ts
@@ -39,6 +39,7 @@ export type IDefaultViewSettingsInit<N extends INodeBase, E extends IEdgeBase> =
 
 export class DefaultView<N extends INodeBase, E extends IEdgeBase> implements IOrbView<IDefaultViewSettings<N, E>> {
   private _container: HTMLElement;
+  private _resizeObs: ResizeObserver;
   private _graph: IGraph<N, E>;
   private _events: OrbEmitter<N, E>;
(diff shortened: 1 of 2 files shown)


tonilastre (approved):
Thanks a lot for the contribution! :)

PR #93, opened 2024-02-28
New: Add properties setters and getters

### LINKED ISSUE #88: Add setters and getters for nodes and edges
Nodes and edges should have setters and getters for properties, data, and positions in the following manner:

```
getX()
patchX({ ... })
patchX((x) => ({ ... }))
setX({ ... })
setX((x) => ({ ... }))
```

The graph model should be able to listen for those changes and know if something has changed, which can be used to call `renderer.render`, a throttled call.

Another part of this task is `states` which is partially done here: https://github.com/memgraph/orb/pull/71

AlexIchenskiy:
Resolves #88

This PR introduces setters, patchers, and getters for node/edge properties, along with implementing the Observer pattern to invoke view rerender upon property changes. The API has been significantly changed. Instead of writing unsafe code like 
```typescript
orb.data.getNodeById(0).property = x;
``` 
we now use 
```typescript
orb.data.getNodeById(0).setProperty(x);
``` 
and 
```typescript
orb.data.getNodeById(0).patchProperty(x);
```

Accessing properties is also modified, from
```typescript 
orb.data.getNodeById(0).x;
```
to 
```typescript 
orb.data.getNodeById(0).getX();
```

Documentation needs to be updated accordingly. However, I believe it's more appropriate to do a complete rewrite upon the 1.0.0 release. Additionally, I want to implement tests after discussing all the details.

This PR is marked as a draft because I feel there are important aspects to double-check and review. However, it's open for all sorts of comments and advice!


AlexIchenskiy:
Tests are failing because of the usage of the `structuredClone()` function, which was introduced in Node.js 17.0.0, while our tests are running on Node.js v16.20.2. Is it ok to update the Node.js version in the `package.json` to >=17.0.0, or should I use some custom function for deep cloning objects?

src/models/observer.ts
```diff
@@ -0,0 +1,11 @@
+export interface IObserver {
+  update(): void;
+}
+
+export interface ISubject {
+  addListener(observer: IObserver): void;
```
AlexIchenskiy, line 6: I'm using the interface `IObserver` along with functions like `addListener`. It sounds natural to me, but I wanted to double-check if this naming convention (using Observable and Listener in the same context) is fine.

src/models/node.ts
```diff
@@ -117,39 +141,73 @@ export const isNode = <N extends INodeBase, E extends IEdgeBase>(obj: any): obj
...
+  private readonly _listeners: IObserver[] = [];
   private readonly _inEdgesById: { [id: number]: IEdge<N, E> } = {};
   private readonly _outEdgesById: { [id: number]: IEdge<N, E> } = {};
   private readonly _onLoadedImage?: () => void;
 
-  constructor(data: INodeData<N>, settings?: Partial<INodeSettings>) {
-    this.id = data.data.id;
-    this.data = data.data;
-    this.position = { id: this.id };
+  constructor(data: INodeData<N>, settings?: Partial<INodeSettings>, listener?: IObserver) {
```
AlexIchenskiy, line 155: Is it fine to use a single listener in the constructor here? From an OOP perspective, using a listener list seems more natural, but I feel that a single listener is clearer and easier to use.

src/models/node.ts
```diff
@@ -363,6 +423,139 @@ export class Node<N extends INodeBase, E extends IEdgeBase> implements INode<N,
...
+  setPosition(position: INodeCoordinates | INodeMapCoordinates | INodePosition): void;
+
+  setPosition(callback: (node: INode<N, E>) => INodeCoordinates | INodeMapCoordinates | INodePosition): void;
+
+  setPosition(
+    arg:
+      | INodeCoordinates
+      | INodeMapCoordinates
+      | INodePosition
+      | ((node: INode<N, E>) => INodeCoordinates | INodeMapCoordinates | INodePosition),
```
AlexIchenskiy, line 483: Should I merge these interfaces somehow? I was considering assigning a type to nodes (for map view and regular view), but I'm unsure if it's necessary since users can use any value from the node data for `lng` and `lat`. Is there a better approach to setting the position for map view nodes maybe?

src/models/edge.ts
```diff
@@ -81,18 +82,20 @@ export enum EdgeType {
...
-  data: E;
-  position: IEdgePosition;
-  style: IEdgeStyle;
-  state: number;
+export interface IEdge<N extends INodeBase, E extends IEdgeBase> extends ISubject {
   readonly id: any;
   readonly offset: number;
   readonly start: any;
   readonly startNode: INode<N, E>;
   readonly end: any;
```
AlexIchenskiy, line 90: I was considering rewriting the values to use getters for consistency, but some are required in the `IEdgeBase` interface (such as `id`, `start`, `end`), which is also required in `topology.ts`. While I believe it's a good idea to use getters for consistency, it seems unnatural given the existing interface requirements. Is it fine to have a combination of TypeScript getters (e.g., `get x`) and regular getters (e.g., `getX`), or should I consider modifying the API further?

src/models/edge.ts
```diff
@@ -129,13 +142,16 @@ export class EdgeFactory {
...
+      data: edge.getData(),
       offset: data?.offset !== undefined ? data.offset : edge.offset,
       startNode: edge.startNode,
       endNode: edge.endNode,
     });
-    newEdge.state = edge.state;
-    newEdge.style = edge.style;
+    newEdge.setState(edge.getState());
+    newEdge.setStyle(edge.getStyle());
+    edge.getListeners().forEach((listener) => {
```
AlexIchenskiy, line 152: It would be nice to make `getListeners()` a `private` or `protected` function, but in that case it can't be used in factory here. Should I leave it public or is there some kind of a workaround?

tonilastre:
> Tests are failing because of the usage of the `structuredClone()` function, which was introduced in Node.js 17.0.0, while our tests are running on Node.js v16.20.2. Is it ok to update the Node.js version in the `package.json` to >=17.0.0, or should I use some custom function for deep cloning objects?

Yep, we should update Node to some newer version for sure!

src/renderer/canvas/node.ts
```diff
@@ -156,34 +157,34 @@ const setupShadow = <N extends INodeBase, E extends IEdgeBase>(
...
-  if (node.style.shadowOffsetX) {
-    context.shadowOffsetX = node.style.shadowOffsetX;
+  if (node.getStyle().shadowOffsetX) {
+    context.shadowOffsetX = node.getStyle().shadowOffsetX as number;
   }
-  if (node.style.shadowOffsetY) {
-    context.shadowOffsetY = node.style.shadowOffsetY;
+  if (node.getStyle().shadowOffsetY) {
+    context.shadowOffsetY = node.getStyle().shadowOffsetY as number;
   }
```
tonilastre, line 171: In these cases, it would be better to get style before checking each condition because right now you are doing 8 deep copies from your node style (that has 30-40 properties). On N number of nodes and render cycles, this will slow down things a lot. This applies to all other usages like this.

src/models/style.ts
```diff
@@ -34,7 +34,7 @@ const getPredefinedLabel = <N extends INodeBase, E extends IEdgeBase>(
   obj: INode<N, E> | IEdge<N, E>,
 ): string | undefined => {
   for (let i = 0; i < LABEL_PROPERTY_NAMES.length; i++) {
-    const value = (obj.data as any)[LABEL_PROPERTY_NAMES[i]];
+    const value = (obj.getData() as any)[LABEL_PROPERTY_NAMES[i]];
```
tonilastre, line 37: The same optimization is required. Get data before the `for loop` because of deep copy.

src/simulator/factory.ts
```diff
@@ -2,6 +2,7 @@ import { ISimulator } from './shared';
 import { MainThreadSimulator } from './types/main-thread-simulator';
 import { WebWorkerSimulator } from './types/web-worker-simulator/web-worker-simulator';
 
+// TODO(dlozic & Alex): CORS handling
```
tonilastre, line 5: You can create an issue for this and then reference an issue here.

src/models/edge.ts
```diff
@@ -293,10 +338,105 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> implements IEdge<N
         return null;
     }
   }
+
+  setData(data: E): void;
+
+  setData(callback: (edge: IEdge<N, E>) => E): void;
+
+  setData(arg: E | ((edge: IEdge<N, E>) => E)) {
```
tonilastre, line 346:
```suggestion
  setData(data: E): void;
  setData(callback: (edge: IEdge<N, E>) => E): void;
  setData(arg: E | ((edge: IEdge<N, E>) => E)) {
```

Maybe it makes sense to group them like this to have a better view of function overloads.

src/models/edge.ts
```diff
@@ -293,10 +338,105 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> implements IEdge<N
...
+    if (typeof arg === 'function') {
+      data = (arg as (edge: IEdge<N, E>) => Partial<E>)(this);
+    } else {
+      data = arg as Partial<E>;
+    }
+
+    Object.keys(data).forEach((key) => {
+      // @ts-ignore
+      this._data[key] = data[key];
+    });
```
tonilastre, line 371: Can you try to do this without `// @ts-ignore`. You can read more about why you need to do that - it is because of `Object.keys` - (try sending an object where keys are numbers and see what happens). JS is a weird thing :D

tonilastre: Also try to skip using `forEach`. It is ok on simple usages like this which won't be called frequently, but `forEach`, even though the code is nicer, it is slower than `for` or `for in` or `for of`.

src/models/edge.ts
```diff
@@ -293,10 +338,105 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> implements IEdge<N
...
+  }
+
+  removeListener(observer: IObserver): void {
+    this._listeners.splice(this._listeners.indexOf(observer), 1);
+  }
+
+  notifyListeners(): void {
+    this._listeners.forEach((listener) => {
+      listener.update();
+    });
```
tonilastre, line 434: Update `forEach` to `for`.

src/models/edge.ts
```diff
@@ -293,10 +338,105 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> implements IEdge<N
...
+    if (typeof arg === 'function') {
+      this._state = (arg as (edge: IEdge<N, E>) => number)(this);
+    } else {
+      this._state = arg as number;
+    }
+    this.notifyListeners();
+  }
+
+  addListener(observer: IObserver): void {
+    this._listeners.push(observer);
```
tonilastre, line 424: Do we want to check if we already have this observer in?

src/models/node.ts
```diff
@@ -363,6 +423,139 @@ export class Node<N extends INodeBase, E extends IEdgeBase> implements INode<N,
...
+  patchData(arg: Partial<N> | ((node: INode<N, E>) => Partial<N>)) {
+    let data: Partial<N>;
+
+    if (typeof arg === 'function') {
+      data = (arg as (node: INode<N, E>) => Partial<N>)(this);
+    } else {
+      data = arg as Partial<N>;
+    }
+
+    Object.keys(data).forEach((key) => {
```
tonilastre, line 466: You can check edge for same comments.

src/models/observer.ts
tonilastre: Yep this is ok, this might just go to some utils or shared stuff - it is not an Orb model, more like utility interface.

src/models/node.ts
tonilastre: Actually, it would be best to add listeners as an array to the `INodeSettings` because those are the initial settings for the node. Having more than > 2 params in JS is making things harder to use (e.g. you want to send a listener with default settings that would look like: `(data, undefined, listener)` 🤮 :D 

You can see that `INodeSettings` already has some utility stuff like callback for onLoadedImage - which in the future should be moved to Listener - but then we need to add what kind of events you want to listen to.

Same applies to edges.

tonilastre: Also having this partial settings make things much more cleaner from "reading the code" perspective:

E.g.
`new Node(data, undefined, this);` - hard to know what `this` is for?

vs

`new Node(data, { listeners: [this] })` - object gives you a key `listeners`.

src/views/orb-map-view.ts
```diff
@@ -82,14 +85,18 @@ export class OrbMapView<N extends INodeBase, E extends IEdgeBase> implements IOr
...
+      {
+        onLoadedImages: () => {
+          // Not to call render() before user's .render()
+          if (this._renderer.isInitiallyRendered) {
+            this.render();
+          }
+        },
       },
-    });
+      this,
```
tonilastre, line 98: What I was saying in one of the comments before :D This is hard to read, you need to go to the object to figure out why `this` and what it is for.

src/views/orb-map-view.ts
```diff
@@ -226,6 +233,10 @@ export class OrbMapView<N extends INodeBase, E extends IEdgeBase> implements IOr
     this._canvas.outerHTML = '';
   }
 
+  update(): void {
+    this.render();
+  }
```
tonilastre, line 238: Hmm, this is maybe a bit tricky, because on external API (orb view), there is an `.update()` call because of an observer. Is there a way to handle this better? Would an observer with a callback do the trick or?

src/models/graph.ts
```diff
@@ -372,12 +378,31 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N
...
+
+  removeListener(observer: IObserver): void {
+    this._listeners.splice(this._listeners.indexOf(observer), 1);
+  }
+
+  notifyListeners(): void {
+    this._listeners.forEach((listener) => {
+      listener.update();
+    });
+  }
```
tonilastre, line 393: This code is repeated (in node/edge) which means it would be great to have this outside as a class implementing IObserver and/or ISubject. 

You can also check how nodejs did it with EventEmitter - it is similar pattern.

tonilastre (commented):
This is going in a great direction!

I still need to check 3 of your comments, so stay tuned for responses.

src/models/edge.ts
```diff
@@ -107,20 +110,30 @@ export interface IEdge<N extends INodeBase, E extends IEdgeBase> {
   getWidth(): number;
   getColor(): Color | string | undefined;
   getLineDashPattern(): number[] | null;
+  setData(data: E): void;
+  setData(callback: (edge: IEdge<N, E>) => E): void;
```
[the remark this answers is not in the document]
AlexIchenskiy: Because I think it is more flexible and provides more parameters to play with. While `data: E` in most cases provides only `id`, `start` and `end` (for an edge), `IEdge` provides more parameters that can be used in the callback, like `position` and `style`.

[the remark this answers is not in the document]
tonilastre: Aha ok. Just FYI, `data: E` can be anything that the user wants, at a minimum it has `id`, `start`, `end` (for an edge). Any data the user wants to have on the node/edge will be here (e.g. Lab saves node properties from the database as data).

src/views/orb-map-view.ts
tonilastre: Is there anything we can do there? Do you have any ideas?

src/utils/observer.utils.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  getListeners(): IObserver[] {
+    return [...this.listeners];
+  }
+
+  removeListener(observer: IObserver): void {
+    if (!this.listeners.includes(observer)) {
+      this.listeners.push(observer);
+    }
+  }
```
tonilastre, line 36: This seems incorrect. If it doesn't have that observer, add it to the list where the name is `removeListener`?

src/utils/object.utils.ts
```diff
@@ -120,3 +120,11 @@ const copyPlainObject = <T>(obj: Record<string, T>): Record<string, T> => {
   });
   return newObject;
 };
+
+export const copyProperties = <T>(source: T, target: T): void => {
```
tonilastre, line 124: Few things here. Maybe better name would be `patch...` rather than `copy...`. When I see `copy` I would expect function to return me copied, not do any changes on the input data.

Also, nitpicking, but ordering matters too - `(target, source)` makes more sense because of `target <= source`. All resulting variables are always on the left side or first one.

src/renderer/canvas/edge/base.ts
```diff
@@ -145,34 +147,38 @@ const setupShadow = <N extends INodeBase, E extends IEdgeBase>(
...
   }
-  if (edge.style.shadowOffsetX) {
-    context.shadowOffsetX = edge.style.shadowOffsetX;
+  if (edgeStyle.shadowOffsetX) {
+    context.shadowOffsetX = edgeStyle.shadowOffsetX as number;
   }
-  if (edge.style.shadowOffsetY) {
-    context.shadowOffsetY = edge.style.shadowOffsetY;
+  if (edgeStyle.shadowOffsetY) {
+    context.shadowOffsetY = edgeStyle.shadowOffsetY as number;
```
cizl, line 162: Why these `as`? I checked in VSCode - aren't those properties already numbers?

If you need strictly numbers, you can use nullish coalescing to get rid of `undefined` -> `edgeStyle.shadowSize = edgeStyle.shadowSize ?? 0`. Maybe then these if conditions aren't necessary then. Not sure if we want to set these values to 0 if they aren't defined.

tonilastre: Good catch for the `as number`. Btw conditions are here to dismiss any unnecessary setup of context variables because even setting `context.shadowOffsetY = 0` will do something. These conditions were here just to setup context only if it is any non-zero number.

src/models/node.ts
```diff
@@ -363,6 +430,112 @@ export class Node<N extends INodeBase, E extends IEdgeBase> implements INode<N,
...
+    callback: (node: INode<N, E>) => INodeCoordinates | INodeMapCoordinates | INodePosition,
+    isInner?: boolean,
+  ): void;
+  setPosition(
+    arg:
+      | INodeCoordinates
+      | INodeMapCoordinates
+      | INodePosition
+      | ((node: INode<N, E>) => INodeCoordinates | INodeMapCoordinates | INodePosition),
+    isInner?: boolean,
```
AlexIchenskiy, line 471: I have to use the second argument here because sometimes we don't need to notify listeners on set position (e.g., on `SIMULATION_PROGRESS` in `orb-view.ts` because position handling is done through the simulation). However, this approach does not seem to be very clean. Do you have any ideas on how I could make it better?

AlexIchenskiy: If we decide to leave it here, perhaps a better naming option would be `isInnerCall`? I will also update the interface accordingly when we define it :)

src/views/orb-map-view.ts
cizl: Do you maybe have an idea how to implement this?

tonilastre: Currently I am seeing two ideas, but there can be more for sure:
1) Have a proxy class that is an actual listener, but also gives you a callback way to respond to updates. Then view can initiate that proxy class where `update` will be called, and update calls a callback that is registered privately from the view. Then `update` is not exposed - easier
2) Add callback to the observer pattern, so listener can either have a callback or extending the `update` to get calls for notifications  - harder

src/models/node.ts
tonilastre: Aha. Angular uses `dispatch: false` for those kind of events. `isInnerCall` doesn't really say what you described so I would go into the direction such as: `isNotifySkipped` or `isDispatchSkipped`, `isNotifyIgnored`. In general, use the name that describes what you described to me, you don't want to notify listeners. And the default should be false or not existing object, that's why I used `skipped` or `Ignored`. Feel free to come up with a name with such constraints.

tonilastre: Also, this should be an object, optional `options?`.

tonilastre (changes requested):

src/views/orb-view.ts
```diff
@@ -282,7 +286,7 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
       for (let i = 0; i < nodes.length; i++) {
         const position = this._settings.getPosition(nodes[i]);
         if (position) {
-          nodes[i].position = { id: nodes[i].id, ...position };
+          nodes[i].setPosition({ id: nodes[i].getId(), ...position }, true);
```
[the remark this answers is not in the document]
cizl: @tonilastre  When passing objects as inputs to functions, how do we handle sane defaults?
When using function arguments we can put some defaults, but when using objects, we need to specify each and every property.

On a call earlier this week we discussed having a default object in that case, and passing a `Partial<IInputParams>` to override the "defaults". Does this make sense? How is this usually handled in practice?

src/views/orb-view.ts
```diff
@@ -594,6 +598,26 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+        nodes: [
+          {
+            x: data.x,
+            y: data.y,
+            sx: data.x,
+            sy: data.y,
+            fx: data.x,
+            fy: data.y,
+            id: data.id,
+          },
```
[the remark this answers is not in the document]
cizl: I think this is fine. I would write a utility function if this starts to get repeated.

But on the other hand, when can this be called? Will it interfere with the physics functionality that depends on `fx` and `fy`? Is the nodes in the `data` always guaranteed to be sticky?

src/views/orb-view.ts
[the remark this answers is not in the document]
tonilastre: It depends on the options. In some other comment, I suggested what the options should be for this one where empty options would mean normal mail flow. Only in exception, the options can be added. 

If that is not the case, then yea, the function should define defaults internally and merge it with the input options.

src/utils/observer.utils.ts
```diff
@@ -0,0 +1,48 @@
+import { INodeCoordinates, INodeMapCoordinates, INodePosition } from '../models/node';
+
+export type IObserverDataPayload = INodePosition | INodeCoordinates | INodeMapCoordinates;
+
+export interface IObserver {
+  update(data?: IObserverDataPayload): void;
+}
```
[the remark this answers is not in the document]
AlexIchenskiy: I am considering adding parameterization for data by including parameters that represent the data that can be sent (similar to this payload but specified in the parameters of each implementation). Do you maybe have any better ideas?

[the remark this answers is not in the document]
AlexIchenskiy: I have created an issue #97 for this where we can discuss it further :)

src/views/orb-map-view.ts
```diff
@@ -89,6 +90,7 @@ export class OrbMapView<N extends INodeBase, E extends IEdgeBase> implements IOr
           this.render();
         }
       },
+      listeners: [this._update],
```
tonilastre, line 93: It would be great to add a comment why this line and not `this`. Same for map view.

tonilastre (approved):

src/views/orb-map-view.ts
AlexIchenskiy: I've added this comment in `observer.utils.ts` so we don't have to add it everywhere. Is it fine, or should I add it to the view files as well?

PR #95, opened 2024-03-11
New: Add state setters with options

### LINKED ISSUE #68: Any way to select only node/edge which is clicked

AlexIchenskiy:
Resolves #68

As an enhancement to the freshly implemented getters and setters, this PR introduces a new way to set the state of nodes and edges using objects that specify state with additional parameters such as `isSingle` and `isToggle`. Now, it is possible to use both the regular setter:
```typescript
node.setState(state)
```
and the new setter:
```typescript
node.setState({ state: GraphObjectState.SELECTED, options: { isSingle: true } })
```
```typescript
edge.setState({ state: GraphObjectState.SELECTED, options: { isSingle: false, isToggle: true } })
```
and so on. This provides more customization flexibility to the user, with default values for these parameters set to `false`. It should be easily extendable for adding even more options.

### DIFF
--- a/src/models/edge.ts
+++ b/src/models/edge.ts
@@ -1,7 +1,7 @@
 import { INodeBase, INode } from './node';
-import { GraphObjectState, IGraphObjectStateParameters } from './state';
+import { GraphObjectState, IGraphObjectStateOptions, IGraphObjectStateParameters } from './state';
 import { Color, IPosition, ICircle, getDistanceToLine } from '../common';
-import { isArrayOfNumbers, isFunction } from '../utils/type.utils';
+import { isArrayOfNumbers, isFunction, isNumber, isPlainObject } from '../utils/type.utils';
 import { IObserver, ISubject, Subject } from '../utils/observer.utils';
 import { patchProperties } from '../utils/object.utils';
 
@@ -420,36 +420,30 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> extends Subject im
       result = arg;
     }
 
-    if (typeof result === 'number') {
+    if (isNumber(result)) {
       this._state = result;
-    } else if (typeof result === 'object') {
+    } else if (isPlainObject(result) && result.options) {
       const options = result.options;
 
-      if (options && options.isToggle) {
-        this._toggleState(result.state);
-      } else {
-        this._state = result.state;
-      }
+      this._state = this._handleState(result.state, options);
 
-      if (options) {
-        this.notifyListeners({
-          id: this.id,
-          type: 'edge',
-          options: options,
-        });
+      this.notifyListeners({
+        id: this.id,
+        type: 'edge',
+        options: options,
+      });
 
-        return;
-      }
+      return;
     }
 
     this.notifyListeners();
   }
 
-  private _toggleState(state: number) {
-    if (this._state === state) {
-      this._state = GraphObjectState.NONE;
+  private _handleState(state: number, options: IGraphObjectStateOptions): number {
+    if (options.isToggle && this._state === state) {
+      return GraphObjectState.NONE;
     } else {
-      this._state = state;
+      return state;
     }
   }
 }
--- a/src/models/edge.ts
+++ b/src/models/edge.ts
@@ -422,25 +422,27 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> extends Subject im
 
     if (isNumber(result)) {
       this._state = result;
-    } else if (isPlainObject(result) && result.options) {
+    } else if (isPlainObject(result)) {
       const options = result.options;
 
       this._state = this._handleState(result.state, options);
 
-      this.notifyListeners({
-        id: this.id,
-        type: 'edge',
-        options: options,
-      });
+      if (options) {
+        this.notifyListeners({
+          id: this.id,
+          type: 'edge',
+          options: options,
+        });
 
-      return;
+        return;
+      }
     }
 
     this.notifyListeners();
   }
 
-  private _handleState(state: number, options: IGraphObjectStateOptions): number {
-    if (options.isToggle && this._state === state) {
+  private _handleState(state: number, options?: Partial<IGraphObjectStateOptions>): number {
+    if (options?.isToggle && this._state === state) {
       return GraphObjectState.NONE;
     } else {
       return state;
--- a/src/models/edge.ts
+++ b/src/models/edge.ts
@@ -1,5 +1,5 @@
 import { INodeBase, INode } from './node';
-import { GraphObjectState } from './state';
+import { GraphObjectState, IGraphObjectStateParameters } from './state';
 import { Color, IPosition, ICircle, getDistanceToLine } from '../common';
 import { isArrayOfNumbers, isFunction } from '../utils/type.utils';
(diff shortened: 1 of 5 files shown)


src/models/graph.ts
```diff
@@ -378,6 +378,27 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> extends Subject imp
   }
 
   update(data?: IObserverDataPayload): void {
+    if (data && 'type' in data && 'options' in data && 'isSingle' in data.options) {
+      if (data.type === 'node' && data.options.isSingle) {
+        const nodes = this._nodes.getAll();
+
+        for (let i = 0; i < nodes.length; i++) {
```
AlexIchenskiy, line 385: I feel like these iterations through all nodes/edges can be slow. Any ideas? And do we want to clear the state of all nodes/edges or only the adjusting ones?

cizl: Or maybe to supplement this nodes array we can keep an index where `nodeIndex[node.id] = nodes.findIndex((el) => el.id === node.id)`

AlexIchenskiy: Are you suggesting storing nodes in a hashmap? As far as I know, retrieving elements from an array by a known index is `O(1)` complexity, so iterating through an array vs a hashmap (which also has `O(1)` complexity for retrieving element by key) may not make a big difference here. I might have misunderstood your point, or perhaps I missed something, so feel free to clarify it further :)

src/models/edge.ts
```diff
@@ -400,15 +402,56 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> extends Subject im
...
+    if (typeof result === 'number') {
+      this._state = result;
+    } else if (typeof result === 'object') {
+      const options = result.options;
+
+      if (options && options.isToggle) {
+        this._toggleState(result.state);
+      } else {
+        this._state = result.state;
+      }
```
tonilastre, line 432: Either have `this._state = <function that handles options in general>(results.state)` where function doesn't mutate the state, but returns what the new state should be, or keep this as is, but still change the `toggleState` to not mutate the `this._state` because I don't see a reason why it should. The best functions are the clean ones, which do not mutate state (if it is possible to make them).

src/models/edge.ts
```diff
@@ -400,15 +402,56 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> extends Subject im
...
+      result = (arg as (edge: IEdge<N, E>) => number | IGraphObjectStateParameters)(this);
     } else {
-      this._state = arg as number;
+      result = arg;
     }
+
+    if (typeof result === 'number') {
+      this._state = result;
+    } else if (typeof result === 'object') {
+      const options = result.options;
```
tonilastre, line 426: For these checks, please use `utils/type.utils.ts` because of:
- cleaner code, `isNumber` vs `typeof x === 'number'
- handles edge cases, `isPlainObject` vs `typeof result === 'object'` (edge case is: `typeof null` will return `'object'`).

src/models/node.ts
```diff
@@ -526,17 +528,58 @@ export class Node<N extends INodeBase, E extends IEdgeBase> extends Subject impl
...
+
     if (isFunction(arg)) {
-      this._state = (arg as (node: INode<N, E>) => number)(this);
+      result = (arg as (node: INode<N, E>) => number | IGraphObjectStateParameters)(this);
     } else {
-      this._state = arg as number;
+      result = arg;
+    }
+
+    if (typeof result === 'number') {
```
tonilastre, line 549: Same comments as for edges.

src/models/state.ts
```diff
@@ -1,6 +1,24 @@
...
 export const GraphObjectState = {
   NONE: 0,
   SELECTED: 1,
   HOVERED: 2,
 };
+
+export interface IGraphObjectStateOptions {
+  isToggle?: boolean;
+  isSingle?: boolean;
+}
```
tonilastre, line 13: In these cases, when it is options, it would be best not to have optionals here, but where you use it, line 17: `Partial<IGraphObjectStateOptions>` where options can also be optional, e.g. `options?: Partial<...>`.

If you move external logic (having optional params) too deep into the type, it is harder to handle that in some other cases. E.g. if you wish to create default params, you could then just say: `const default: IGraphObjectStateOptions = { ... }` and you would be forced to define both. If you would use `options` from `Parameters` you would see that this object is optional here with all its children. The context of when is something optional, when it isn't is being kept.

src/models/graph.ts
tonilastre: If all nodes are in state X, and you want to clear them up if they are not node with ID 1, you still need to to `N - 1` updates. So hash maps or any other structure won't help there, you already have the fastest structure, an array.

But, you can do this and increase the memory footprint: I am not sure if it makes sense to go into this right now, but this would be the plan:
- Graph keeps a structure of nodes/edges by state, so each state has a structure of nodes and edges (probably just id) - a set of a map.
- States in nodes and edges are just views, readonly
- When state is updated in node, edge it is just updated in the graph

This way, if you have `isSingle: true`, graph, as an owner of the data can then literally take all selected nodes from a set or a map, clear those states and set the new node as single selected. This would improve the speed, but it would worse the memory footprint and code simplitiy because owner of the state becomes the graph, not graph objects (node/edge).

I would keep this as is and fix only if we see that it hurts performance.

src/models/graph.ts
```diff
@@ -378,6 +378,27 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> extends Subject imp
   }
 
   update(data?: IObserverDataPayload): void {
+    if (data && 'type' in data && 'options' in data && 'isSingle' in data.options) {
+      if (data.type === 'node' && data.options.isSingle) {
```
tonilastre, line 382: Can you maybe simplify this to `if (data && data.options?.isSingle && data.type === 'node')`?

tonilastre (changes requested):

src/models/graph.ts
AlexIchenskiy: I believe it could be possible, but only after reworking `IObserverDataPayload` into a more generic solution as you mentioned [here](https://github.com/memgraph/orb/pull/93#discussion_r1523676765). Currently, I am using some sort of custom type guard because of the interfaces used in that payload

src/models/graph.ts
cizl: Sorry, I misunderstood the functionality. I saw `===` instead of `!==`. I thought you were trying to clear the state of anode with a specific `id`.

src/models/edge.ts
```diff
@@ -400,15 +402,52 @@ abstract class Edge<N extends INodeBase, E extends IEdgeBase> extends Subject im
...
+        });
+
+        return;
+      }
     }
+
     this.notifyListeners();
   }
+
+  private _handleState(state: number, options?: Partial<IGraphObjectStateOptions>): number {
```
tonilastre, line 444: It is nitpicking really, but maybe `getNextState` would be a better name.

tonilastre (approved):

PR #96, opened 2024-03-13
New: Add removal functions

AlexIchenskiy:
This PR introduces functions for removing all data, nodes (the same functionality as removing all the data), or edges by reusing existing remove function. It can be used as follows:
```typescript
orb.data.removeAll();
orb.data.removeAllNodes();
orb.data.removeAllEdges();
```

### DIFF
--- a/src/models/graph.ts
+++ b/src/models/graph.ts
@@ -279,15 +279,15 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N
     const removedNodesData = this._removeNodes(nodeIds);
     const removedEdgesData = this._removeEdges(edgeIds);
 
-    const removedData = {
-      nodeIds: dedupArrays(removedNodesData.nodeIds, removedEdgesData.nodeIds),
-      edgeIds: dedupArrays(removedNodesData.edgeIds, removedEdgesData.edgeIds),
-    };
-
     this._applyEdgeOffsets();
     this._applyStyle();
 
     if (this._settings && this._settings.onRemoveData) {
+      const removedData: IGraphObjectsIds = {
+        nodeIds: dedupArrays(removedNodesData.nodeIds, removedEdgesData.nodeIds),
+        edgeIds: dedupArrays(removedNodesData.edgeIds, removedEdgesData.edgeIds),
+      };
+
       this._settings.onRemoveData(removedData);
     }
   }
--- a/src/models/graph.ts
+++ b/src/models/graph.ts
@@ -47,7 +47,7 @@ export interface IGraphSettings<N extends INodeBase, E extends IEdgeBase> {
   onLoadedImages?: () => void;
   onSetupData?: (data: Partial<IGraphData<N, E>>) => void;
   onMergeData?: (data: Partial<IGraphData<N, E>>) => void;
-  onRemoveData?: (data: Partial<{ nodeIds: number[]; edgeIds: number[] }>) => void;
+  onRemoveData?: (data: Partial<{ nodeIds: any[]; edgeIds: any[] }>) => void;
 }
 
 export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N, E> {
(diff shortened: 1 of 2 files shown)


src/models/graph.ts
```diff
@@ -274,7 +273,35 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N
     this._applyEdgeOffsets();
     this._applyStyle();
 
-    this._settings?.onRemoveData?.(data);
+    if (this._settings && this._settings.onRemoveData) {
+      this._settings.onRemoveData(data);
```
tonilastre, line 277: As a user (internal or external) I would expect to receive in the callback all nodes and edges that have been removed. You are forwarding the data that you've received which doesn't necessarily mean that those have been removed, e.g.

In my graph I have these nodes and edges
- Nodes: 1, 2, 3
- Edges A (1-2), B (2-1), C(1-3)

I call remove with `nodes = 1, 3, 5` and `edges = C, D`. But actually nodes `1, 3` will be removed and all edges will be removed (`A, B, C`). So instead of receiving `[1, 3, 5], [C, D]` in the callback, I would expect to receive `[1, 3], [A, B, C]`.

tonilastre (changes requested):
Overall looks great! There is just one comment on the callback data.

src/models/graph.ts
```diff
@@ -263,18 +263,41 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N
...
 
-    this._removeNodes(nodeIds);
-    this._removeEdges(edgeIds);
+    const removedData = this._removeNodes(nodeIds);
+    const removedEdgeIds = this._removeEdges(edgeIds);
+
+    // Merge edges removed by removing nodes and by removing edges, ensuring there are no duplicate edge IDs.
+    removedData.edgeIds = removedData.edgeIds.concat(
+      removedEdgeIds.filter((edgeId) => removedData.edgeIds.indexOf(edgeId) < 0),
+    );
```
AlexIchenskiy, line 276: This is the most elegant solution I've found so far, but I'm open to suggestions for any more performant/readable alternatives

src/models/graph.ts
```diff
@@ -474,7 +497,7 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N
     this._edges.removeMany(removedEdgeIds);
   }
 
-  private _removeNodes(nodeIds: any[]) {
+  private _removeNodes(nodeIds: any[]): { nodeIds: any[]; edgeIds: any[] } {
```
tonilastre, line 500: This return type is cool. I would maybe extract it into an interface/type. Also, just for a sake of consistency, I would maybe consider returning the same type on `_removeEdges` - right now, nodeIds will be always empty, but in the future if we add "removeConnectingNodes` for edges, then the API still holds and can use the returned object with both `nodeIds` and `edgeIds`.

src/models/graph.ts
tonilastre: This can be slow, so I would suggest creating utility function, e.g. `dedupArrays` that receive a generic array, or multiple arrays and dedups their values - using set or map.

E.g.
```
dedupArrays([1, 2, 3], [1, 2]) -> [1, 2, 3])
dedupArrays([1, 2, 3], [1, 2], [5, 6]) -> [1, 2, 3, 5, 6])
dedupArrays([1, 2, 3], [1, 2], [5, 6], [], [6, 7]) -> [1, 2, 3, 5, 6, 7])
```

tonilastre (changes requested):

src/models/graph.ts
```diff
@@ -263,18 +272,41 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> implements IGraph<N
...
 
-    this._removeNodes(nodeIds);
-    this._removeEdges(edgeIds);
+    const removedNodesData = this._removeNodes(nodeIds);
+    const removedEdgesData = this._removeEdges(edgeIds);
+
+    const removedData = {
+      nodeIds: dedupArrays(removedNodesData.nodeIds, removedEdgesData.nodeIds),
+      edgeIds: dedupArrays(removedNodesData.edgeIds, removedEdgesData.edgeIds),
+    };
```
tonilastre, line 285: Just a small detail here. Whenever you create a new object or an array, always add a type. Types will help you to raise warning if you for example change type in the future.

```
const removedData: IGraphObjectsIds = {
...
```

Also, this object should be within `if` on line 290 because you want to create this object only if settings exists and `settings.onRemoveData` exists, otherwise time and memory is wasted if settings is not called at all.

tonilastre (approved):
Looks good! I've added a small detail for a better code quality.

PR #101, opened 2024-04-26
Docs: Fixed wrong usage

stefanak-michal:


tonilastre (approved):
Great catch! Thanks a lot for fixing it!

PR #106, opened 2025-03-07
Update workflows

jsoref:
* Most github actions/* are running `@v4` and have various performance improvements. A number of the `@v1` releases have been deprecated. I'm actually surprised checkout isn't screaming...
* The release workflow runs code that expects to be running from the memgraph/orb repository and triggers failures otherwise (this happens whenever someone updates the `main` branch in their forks, as in https://github.com/check-spelling-sandbox/orb/actions/runs/13721605401/job/38378121098#step:7:104)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,8 +9,8 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
         with:
           node-version: "16.x"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -8,9 +8,10 @@ on:
 jobs:
   publish:
     runs-on: ubuntu-latest
+    if: github.repository == 'memgraph/orb'
```
tonilastre, line 11: Nice!

tonilastre (approved):
Great PR. Thanks for updating the workflows.

I think it wasn't screaming because it was working with deprecation notices, but starting end of Jan/start of Feb those actions were dismissed :)

PR #107, opened 2025-05-15
New: Add tree layout

AlexIchenskiy:


src/models/graph.ts
```diff
@@ -92,6 +95,22 @@ export class Graph<N extends INodeBase, E extends IEdgeBase> extends Subject imp
     this.notifyListeners();
   }
 
+  setLayout(layout: ILayout<N, E>): void {
+    this._layout = layout;
+    console.log('Layout set', layout);
```
tonilastre: Check for these after debugging is done :D

src/simulator/layout/layout.ts
```diff
@@ -1,29 +1,41 @@
...
+
+export type LayoutSettingsMap = {
+  hierarchical: IHierarchicalLayoutOptions;
+  circular: ICircularLayoutOptions;
+  grid: Record<string, never>;
+};
+
+export interface ILayoutSettings {
+  type: LayoutType;
+  options?: LayoutSettingsMap[LayoutType];
```
tonilastre, line 19: Cool stuff. Do you get an error if you add a new LayoutType that is not referenced in LayoutSettingsMap?

Also, what we talked about: Adding force layout here along with all params (gaps, etc.)

src/simulator/layout/layout.ts
```diff
@@ -1,29 +1,41 @@
...
-  constructor(layoutName: string) {
-    this._layout = this.layoutByLayoutName[layoutName];
-  }
-
-  getPositions(nodes: INode<N, E>[], width: number, height: number): INodePosition[] {
-    return this._layout === null ? [] : this._layout.getPositions(nodes, width, height);
+export class LayoutFactory {
+  static create<N extends INodeBase, E extends IEdgeBase>(
+    width: number,
+    height: number,
```
tonilastre: It would be great to remove dependency on these.

src/simulator/layout/layout.ts
```diff
@@ -1,29 +1,41 @@
...
+  ): ILayout<N, E> | null {
+    switch (settings?.type) {
+      case 'hierarchical':
+        return new HierarchicalLayout<N, E>(width, height, settings.options as IHierarchicalLayoutOptions);
+      case 'circular':
+        return new CircularLayout<N, E>(width, height, settings.options as ICircularLayoutOptions);
+      case 'grid':
+        return new GridLayout<N, E>(width, height);
+      default:
+        return null;
```
tonilastre: Does it make sense to throw an error rather than null? Especially when we will have force layout here.

src/simulator/layout/layouts/grid.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  getPositions(nodes: INode<N, E>[]): INodePosition[] {
+    const rows = Math.ceil(Math.sqrt(nodes.length));
+    const cols = Math.ceil(nodes.length / rows);
+    const cellWidth = this.width / cols;
+    const cellHeight = this.height / rows;
+
+    const positions: INodePosition[] = [];
+    nodes.forEach((node, index) => {
```
tonilastre: Would a grid layout create such a layout where there is a minimum number of edge crosses? Would that be the best grid layout?

src/simulator/layout/layouts/hierarchical.ts
```diff
@@ -0,0 +1,84 @@
...
+import { ILayout } from '../layout';
+
+export type HierarchicalLayoutOrientation = 'horizontal' | 'vertical';
+
+export interface IHierarchicalLayoutOptions {
+  orientation?: HierarchicalLayoutOrientation;
+  reversed?: boolean;
+}
+
+export class HierarchicalLayout<N extends INodeBase, E extends IEdgeBase> implements ILayout<N, E> {
```
tonilastre, line 23: For the HierarchicalLayout, here are some edge cases to cover:

- 5 disconnected nodes - it should be a single row or single column
- 2+ graphs - there could be more root nodes
- A graph with 5 layers, where in each layer there are more nodes - should we minimize edge crosses here?
- Recurring edges, e.g. `A -> B, A -> C, B -> A, A -> A`

tonilastre: Btw, what would happen if you add new nodes? What happens with existing nodes - those might change position? Or if you remove some nodes? (we have that in the Lab with collapse, expand and hide)

src/views/orb-view.ts
```diff
@@ -34,6 +35,7 @@ export interface IOrbViewSettings<N extends INodeBase, E extends IEdgeBase> {
   render: Partial<IRendererSettings>;
   strategy: Partial<IEventStrategySettings>;
   interaction: Partial<IGraphInteractionSettings>;
+  layout: Partial<ILayoutSettings>;
```
tonilastre, line 44: Great setup and field definition

tonilastre (commented):
I did a high level overview of the feature. I will dig deeper into the logic and how it works once these high level stuff is checked and/or completed.

src/simulator/layout/layout.ts
AlexIchenskiy: Yes! I am now adding a force layout, and I do get an error that there is no force layout in the `LayoutSettingsMap`, so it is future-proof for adding new ones :)

AlexIchenskiy: Also, we have force simulation parameters in `SimulatorEngineSettings` (e.g. centering, link distance) - do we want to move them to layout settings?

src/simulator/layout/layouts/grid.ts
AlexIchenskiy: Currently, it just places nodes in the order they were in the list into a grid, but it can be modified later on :)

src/simulator/layout/layout.ts
AlexIchenskiy: Now it throws an error if the layout type is unrecognizable and uses `null` for the force layout because the actual 'layouting' is happening in the simulator. What do you think?

src/simulator/layout/layouts/hierarchical.ts
AlexIchenskiy: In general, root is selected like this: 
1. Pick any node with no incoming edges
2. If no such nodes exist, select the node with the least incoming edges, not counting the recurring edges.
It looks like this for `A -> B, A -> C, B -> A, A -> A`:
image

src/views/orb-view.ts
```diff
@@ -149,37 +161,59 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+          y: _options.centerY,
+        };
+
+        this._settings.simulation.links = {
+          ...(DEFAULT_SETTINGS.links as Required<ID3SimulatorEngineSettingsLinks>),
+          ...this._settings.simulation.links,
+          distance: _options.nodeDistance,
+        };
+      }
+      this._simulator.setSettings(this._settings.simulation);
```
AlexIchenskiy: Here, some simulator settings are overwritten by the same settings but from the layout settings. This ensures back-compatibility, so no existing code breaks, but it also adds a more convenient way to define settings that are logically more for a layout. My only concern here is that this 'adapter' is not as elegant as it could be. Do you have any ideas?

AlexIchenskiy:
I'll refactor the documentation as soon as the logic is polished and finalized 😌

src/simulator/layout/layouts/circular.ts
```diff
@@ -0,0 +1,42 @@
...
+import { INode, INodeBase, INodePosition } from '../../../models/node';
+import { ILayout } from '../layout';
+
+export interface ICircularLayoutOptions {
+  radius?: number;
+  centerX?: number;
+  centerY?: number;
+}
+
+export const DEFAULT_CIRCULAR_LAYOUT_OPTIONS: ICircularLayoutOptions = {
```
tonilastre:
```suggestion
export const DEFAULT_CIRCULAR_LAYOUT_OPTIONS: Required<ICircularLayoutOptions> = {
```

src/simulator/layout/layouts/circular.ts
```diff
@@ -0,0 +1,42 @@
...
+  radius: 100,
```
tonilastre, line 12: Is this maybe midRadius? Because if you have radius 10, but there are 1000000 nodes. The nodes will overlap.

src/simulator/layout/layouts/grid.ts
```diff
@@ -0,0 +1,40 @@
...
+    const rows = Math.ceil(Math.sqrt(nodes.length));
+    const cols = Math.ceil(nodes.length / rows);
+
+    const positions: INodePosition[] = [];
+    nodes.forEach((node, index) => {
+      const row = Math.floor(index / cols);
+      const col = index % cols;
+      const x = col * this._colGap;
+      const y = row * this._rowGap;
+      positions.push({ id: node.getId(), x, y });
```
tonilastre: This can be faster. If you know how big the array will be, then create positions like this:

```
const positions: INodePosition[] = new Array<INodePosition>(size);

positions[i] = ...
```

+ `forEach` -> `for`. because this is called a lot of times.

src/simulator/layout/layouts/hierarchical.ts
```diff
@@ -0,0 +1,160 @@
...
+
+export interface IHierarchicalLayoutOptions {
+  nodeGap?: number;
+  levelGap?: number;
+  treeGap?: number;
+  orientation?: HierarchicalLayoutOrientation;
+  reversed?: boolean;
+}
+
+export const DEFAULT_HIERARCHICAL_LAYOUT_OPTIONS: IHierarchicalLayoutOptions = {
```
tonilastre: It would be better if you add `Required<IHierarchicalLayoutOptions>` to force default definition.

src/simulator/layout/layouts/hierarchical.ts
tonilastre: Yep this is great!

src/simulator/layout/layouts/hierarchical.ts
```diff
@@ -0,0 +1,160 @@
...
+  orientation: 'vertical',
+  reversed: false,
+};
+
+export class HierarchicalLayout<N extends INodeBase, E extends IEdgeBase> implements ILayout<N, E> {
+  private _nodeGap: number;
+  private _levelGap: number;
+  private _treeGap: number;
+  private _orientation: HierarchicalLayoutOrientation;
+  private _reversed: boolean;
```
tonilastre: Is there a reason why all these are separate and not part of some common key, e.g. `config`. Adding new values will be easier because the new value will probably be part of that config?

src/simulator/layout/layouts/hierarchical.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    return positions;
+  }
+
+  getConnectedComponents = (nodes: INode<N, E>[]): INode<N, E>[][] => {
+    const visited = new Set<INode<N, E>>();
+    const components: INode<N, E>[][] = [];
+
+    for (const node of nodes) {
+      if (!visited.has(node)) {
```
tonilastre: Maybe inverse with `continue` and having the rest of the code with -1 indent.

src/simulator/layout/layouts/hierarchical.ts
```diff
@@ -0,0 +1,160 @@
...
+        const component: INode<N, E>[] = [];
+        const queue: INode<N, E>[] = [node];
+        visited.add(node);
+
+        while (queue.length > 0) {
+          const current = queue.pop();
+
+          if (current) {
```
tonilastre: Same here, with inverse if.

src/simulator/layout/layouts/hierarchical.ts
```diff
@@ -0,0 +1,160 @@
...
+    this._nodeGap = _options.nodeGap;
+    this._levelGap = _options.levelGap;
+    this._treeGap = _options.treeGap;
+    this._orientation = _options.orientation;
+    this._reversed = _options.reversed;
+  }
+
+  getPositions(nodes: INode<N, E>[]): INodePosition[] {
+    const components = this.getConnectedComponents(nodes);
+    const positions: INodePosition[] = [];
```
tonilastre: Feel free to create array with all positions for faster assignment, `new Array(size)`

src/simulator/layout/layouts/force.ts
```diff
@@ -0,0 +1,11 @@
+export interface IForceLayoutOptions {
+  centerX?: number;
+  centerY?: number;
+  nodeDistance?: number;
```
tonilastre, line 4: Btw what is with the force config that we have in the Lab: 2 other forces along with link distance (node distance).

tonilastre (changes requested):
Looks great! I've added some comments regarding performance.

Btw there is one issue with the layout change:

If I start with force layout, but change the layout after few seconds to tree layout - it all works fine. Btw maybe we should recenter after layout change by default.

Issue: If I start with tree layout and I want to change to force layout after few seconds, it is not working:

```
    setTimeout(() => {
      orb.setSettings({
        layout: {
          type: 'force',
        },
      });
    }, 3000);
```

src/views/orb-view.ts
tonilastre: Yep it is hard due to back-compatibility. I think this is fine!

src/views/orb-view.ts
```diff
@@ -604,6 +639,72 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+      this.render();
+    });
+    this._simulator.on(SimulatorEventType.SETTINGS_UPDATE, (data) => {
+      this._settings.simulation = data.settings;
+    });
+
+    this._simulator.activateSimulation();
+  };
+
+  private _disableSimulation = () => {
```
AlexIchenskiy, line 683: For enabling/disabling, does it make sense to create some sort of map to store event types and their functions, and iterate through it to enable/disable event listeners? It may be overcomplicating, but it results in less code duplication. What do you think?

tonilastre: That's how EventEmitter actually works. It has `.on` and `.off` and it keeps a map of event names and functions. The thing that I am not sure how it figures out function equality when doing `.off`. https://nodejs.org/en/learn/asynchronous-work/the-nodejs-event-emitter

tonilastre: This doesn't mean that simulator is EventEmitter, but if that path is taken, it would be great to do it in EventEmitter style.

src/simulator/layout/layout.ts
tonilastre: I think it would make sense, but not required

tonilastre (approved):

PR #108, opened 2026-03-04
Fix: Change layout engine logic

AlexIchenskiy:
This PR aims to refactor the core simulation and position calculation logic, fix some long-standing bugs - such as issues with render and recenter callback functionality - and add long-awaited features like simulation cancellation.

List of the most important changes:
- Restructured layout engines into `engines/dynamic/` and `engines/static/` subdirectories with a shared `BaseLayoutEngine` abstract class that consolidates common fields (`_nodes`, `_edges`, `_nodeIndexByNodeId`), cancellation support (`_cancelSimulation`), and a `MessageChannel`-based scheduler (`_scheduleNext`) for non-blocking chunked execution

- Added chunked asynchronous execution to static layout engines (circular, grid, hierarchical) with cancellation support, `_pendingRecalculation` queuing to prevent batch operations from being silently dropped, and `_emitProgress` for progress tracking

- Added chunked asynchronous execution to the force layout engine's `_runSimulation`, replacing the synchronous blocking loop that froze the main thread during simulation

- Changed the `alpha` parameter to drop number of iterations per simulation from ~300 to ~100, improving performance

- Removed the separate `SimulatorSettings` abstraction (`IForceLayoutSettings`) and folded its options into `IForceLayoutOptions`, collapsing `setSettings()` and `setLayoutEngine()` into a single `setSettings(ILayoutSettings)` call - establishing layout settings as the single abstraction layer for all position-recalculation configuration

- Fixed render callback timing - `render(onRendered)` now fires the callback after `SIMULATION_END` + `RENDER_END` instead of synchronously before the simulation completes.

- Fixed transition callbacks - `recenter()`, `zoomIn()`, and `zoomOut()` now use `.on('end', ...)` instead of `.call()` for post-transition callbacks

- Fixed the throttle function - rewrote it with proper time-based gating to prevent stale closures and frozen renders

- Cleaned up the D3 simulation lifecycle - `_resetSimulation` now properly stops and detaches listeners before creating a new simulation

- Simplified and improved the web worker - replaced the verbose switch/case with a handler map, removed the `SetLayoutEngine` message type, created the engine lazily on the first `SetSettings`, and enabled real-time engine switching using the `SetSettings` message

- Designed an `IGraphInteraction` interface to cleanly separate entity interaction concerns (select/unselect, hover/unhover by ID) from the core graph data model, exposed via `view.interaction.*` - consistent with the existing `view.data.*` pattern

- Fixed the `recenter()` logic to accept options so that it recenters correctly for edge cases such as unbalanced tree views


src/simulator/shared.ts
```diff
@@ -74,8 +74,9 @@ export interface ISimulator extends IEmitter<SimulatorEvents> {
   releaseNodes(nodes?: ISimulationNode[]): void;
 
   // Settings handlers
-  setSettings(settings: ID3SimulatorEngineSettingsUpdate): void;
+  setSettings(settings: DeepPartial<ILayoutSettings>): void;
 
+  getIsSimulationRunning(): boolean;
```
tonilastre: Nitpicking, but can this be `isSimulationRunning()`?

src/simulator/types/web-worker-simulator/web-worker-simulator.ts
```diff
@@ -132,8 +134,15 @@ export class WebWorkerSimulator extends Emitter<SimulatorEvents> implements ISim
     this.emitToWorker({ type: WorkerInputType.ReleaseNodes, data: { nodes } });
   }
 
-  setSettings(settings: ID3SimulatorEngineSettingsUpdate) {
-    this.emitToWorker({ type: WorkerInputType.SetSettings, data: settings });
+  setSettings(settings: ILayoutSettings) {
+    this.emitToWorker({
+      type: WorkerInputType.SetSettings,
+      data: settings,
+    } as IWorkerInputPayload);
```
tonilastre: If we have new typescript in this project, you can use `satisfies` here instead of `as`.

src/models/interaction.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface IGraphInteraction {
+  selectNodeById(id: any): boolean;
+  selectEdgeById(id: any): boolean;
+  unselectAll(): number;
+  hoverNodeById(id: any): boolean;
+  hoverEdgeById(id: any): boolean;
+  unhoverAll(): number;
+}
+
+export class GraphInteraction<N extends INodeBase, E extends IEdgeBase> implements IGraphInteraction {
```
tonilastre, line 15: This is cool!

src/simulator/engine/engines/dynamic/force-layout-engine.ts
```diff
@@ -0,0 +1,484 @@
...
+          this._nodeIndexByNodeId[nodeId] = index;
+          this._nodes[index] = data.nodes[i];
+        } else {
+          this._nodes.push(data.nodes[i]);
+        }
+      }
+    }
+
+    if (data.edges) {
+      this._edges = this._edges.concat(data.edges);
```
tonilastre: Why edges are not being merged by id?

src/simulator/engine/engines/dynamic/force-layout-engine.ts
```diff
@@ -0,0 +1,484 @@
...
+    this._isStabilizing = true;
+    this._simulation.alpha(this._settings.alpha.alpha).alphaTarget(this._settings.alpha.alphaTarget).stop();
+
+    const totalSimulationSteps = Math.min(
+      MAX_SIMULATION_STEPS,
+      Math.ceil(Math.log(this._settings.alpha.alphaMin) / Math.log(1 - this._settings.alpha.alphaDecay)),
+    );
+
+    let lastProgress = -1;
+    let i = 0;
```
tonilastre: Nitpicking, but does it make sense to call this `step` instead of `i` because `i` is usually used a blocked locally variable, this `i` is all around in `runChunk`.

src/simulator/engine/engines/static/circular-layout-engine.ts
```diff
@@ -0,0 +1,49 @@
...
+  protected calculatePositions(
+    nodes: ISimulationNode[],
+    _edges: ISimulationEdge[],
+    onProgress: (progress: number) => void,
+    isCancelled: () => boolean,
+    onComplete: () => void,
+  ) {
+    const angleStep = (2 * Math.PI) / nodes.length;
+    let lastProgress = -1;
+    let i = 0;
```
tonilastre: Same comment as before for `i`

src/simulator/engine/engines/static/grid-layout-engine.ts
```diff
@@ -0,0 +1,53 @@
...
+    nodes: ISimulationNode[],
+    _edges: ISimulationEdge[],
+    onProgress: (progress: number) => void,
+    isCancelled: () => boolean,
+    onComplete: () => void,
+  ) {
+    const rows = Math.ceil(Math.sqrt(nodes.length));
+    const cols = Math.ceil(nodes.length / rows);
+    let lastProgress = -1;
+    let i = 0;
```
tonilastre: And here :), and probably on other engines

src/simulator/engine/shared.ts
```diff
@@ -0,0 +1,195 @@
...
+};
+
+export const DEFAULT_FORCE_LAYOUT_OPTIONS: IForceLayoutOptions = {
+  isSimulatingOnDataUpdate: true,
+  isSimulatingOnSettingsUpdate: true,
+  isSimulatingOnUnstick: true,
+  isPhysicsEnabled: false,
+  alpha: {
+    alpha: 1,
+    alphaMin: 0.05, // default alphaMin is 0.001, which results in 285 ticks to converge. Using 0.05 converges to similar stable results in 106 ticks
```
tonilastre, line 53: 👍

src/views/orb-view.ts
```diff
@@ -326,21 +278,38 @@ export class OrbView<N extends INodeBase, E extends IEdgeBase> implements IOrbVi
...
+    const layout = this._settings.layout;
+    const isHorizontal =
+      layout.type === 'hierarchical' && (layout.options as IHierarchicalLayoutOptions).orientation === 'horizontal';
+    const isVertical =
+      layout.type === 'hierarchical' && (layout.options as IHierarchicalLayoutOptions).orientation === 'vertical';
+    const reversed = (layout.options as IHierarchicalLayoutOptions).reversed;
+    const recenterOptions: IFitZoomTransformOptions = {
+      anchorX: isHorizontal ? (reversed ? 'end' : 'start') : 'center',
+      anchorY: isVertical ? (reversed ? 'end' : 'start') : 'center',
+    };
```
tonilastre: Does it make sense to get these anchor variables from the `recenter` itself? We could have defaults for each layout, but user can override it through the argument of this function, e.g.

```
recenter({ anchorX, anchorY }, onRendered?);
recenter(onRendered?);
...
```

tonilastre: Maybe we could also save these defaults for layouts in layout settings too. I find it odd to have these checks here.

tonilastre (approved):
This is great! The structure or layouts and engines are well done, and easy to extend in the future. And the performance, 🚀 !

I also like the graph interaction part because it was always hard to maintain those states from internal node/edge functions.

I've added some comments, mostly nitpicking.

src/views/orb-view.ts
AlexIchenskiy: Yeah, I was skeptical about these checks too - it gets complicated because it depends both on the layout type and the layout options (it's different for horizontal/vertical tree, for example). I've added them to the layout settings, but I also extracted the logic for the default anchoring depending on the current layout; I believe this is the cleanest solution for now

PR #109, opened 2026-06-16
New: Add WebGL renderer

AlexIchenskiy:
This PR introduces an alternative WebGL renderer alongside the existing Canvas renderer, plus a refactor of the layout/simulator engine to support it. The Canvas renderer remains the default, WebGL is fully optional.

The goal is better rendering performance on larger graphs by offloading node/edge/label drawing to the GPU via custom GLSL shaders, and optionally running force-directed layout on the GPU as well.

### What's included

- **WebGL renderer** (`RendererType.WEBGL`) — selectable per view, drawn with custom GLSL shaders for nodes, edges, and labels (`src/renderer/webgl/`).
- **Node images & labels on the GPU** — image atlas (`image-atlas.ts`) and label cache (`label-cache.ts`) so textures/text can be batched.
- **Improved node/shape geometry options** for the WebGL path.
- **GPU force layout engine** (`gpu-force-layout-engine.ts`) — runs the force simulation in WebGL2 with a quadtree/Barnes-Hut approximation. Automatically falls back to the CPU force engine if WebGL2 is unavailable.
- **Layout engine refactor** — the old `d3-simulator-engine` is replaced by a set of pluggable engines (`force`, `circular`, `grid`, `hierarchical`) behind a factory.

### Documentation

Documentation for the WebGL renderer and the new layout engines is still lacking - the APIs may still change while this stabilizes, so docs will be filled in once everything is done and confirmed.

### Usage

Opt into the WebGL renderer when creating a view:

```js
const orb = new Orb.OrbView(container, {
  render: { type: 'webgl' },
});
```

Opt into GPU layout:
```js
const orb = new Orb.OrbView(container, {
  render: { type: 'webgl' },
  layout: { type: 'force', options: { useGPU: true } },
});
```

### Note on stability

This adds a WebGL rendering path, which should be considered experimental and potentially fragile. It depends on GPU/driver behavior and WebGL2 support, so rendering may differ across browsers, GPUs, parameters, and device pixel ratios, and can fail, look different or require cherry-picking for stable results where the Canvas renderer would not. The Canvas renderer remains the stable default and is unchanged in behavior.


src/utils/program.utils.ts
```diff
@@ -0,0 +1,31 @@
...
+  gl: WebGL2RenderingContext,
+  vertexSource: string,
+  fragmentSource: string,
+): WebGLProgram => {
+  const vertexShader = compileShader(gl, vertexSource, ShaderType.VERTEX);
+  const fragmentShader = compileShader(gl, fragmentSource, ShaderType.FRAGMENT);
+
+  const program = gl.createProgram();
+  if (!program) {
+    throw new OrbError('Failed to create program.');
```
tonilastre:
```suggestion
    throw new OrbError('Failed to create GL program.');
```
Having only `program` seems too generic. Same for the error below.

src/renderer/webgl/utils/image-atlas.ts
```diff
@@ -0,0 +1,180 @@
...
+export class ImageAtlas {
+  private _gl: WebGL2RenderingContext;
+  private _canvas: HTMLCanvasElement;
+  private _ctx: CanvasRenderingContext2D;
+  private _texture: WebGLTexture | null = null;
+  private _cache = new Map<string, ImageAtlasEntry>();
+  private _pending = new Map<string, PendingImage>();
+  private _shelves: Shelf[] = [];
+  private _dirty = false;
+  private _textureAllocated = false;
```
tonilastre: Nitpicking, but do want these to start with `_is...`

src/renderer/webgl/utils/color.utils.ts
tonilastre, line 1: Btw do you need this file because it is empty?

src/renderer/webgl/utils/label-cache.ts
```diff
@@ -0,0 +1,181 @@
...
+export interface LabelAtlasEntry {
+  u0: number;
+  v0: number;
+  u1: number;
+  v1: number;
+  pxWidth: number;
+  pxHeight: number;
+}
+
+interface Shelf {
```
tonilastre: `I`

src/renderer/webgl/utils/label-cache.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export class LabelCache {
+  private _gl: WebGL2RenderingContext;
+  private _canvas: HTMLCanvasElement;
+  private _ctx: CanvasRenderingContext2D;
+  private _texture: WebGLTexture | null = null;
+  private _cache = new Map<string, LabelAtlasEntry>();
+  private _shelves: Shelf[] = [];
+  private _dirty = false;
+  private _textureAllocated = false;
```
tonilastre: Similar comment as with previous class about boolean vars

src/simulator/engine/engines/dynamic/gpu-force-layout-engine.ts
```diff
@@ -0,0 +1,1021 @@
...
+    if (this._settings.isSimulatingOnUnstick && this._nodes.length > 0) {
+      this.activateSimulation();
+    }
+  }
+
+  terminate(): void {
+    super.terminate();
+
+    const gl = this._gl;
+    if (gl) {
```
tonilastre: Maybe an early return here.

